PR #6, opened 2016-03-17
Possible bug for logical right shift

sebbsn:
When shifting a signed integer n, where n<0, by 31 bits the current return value would result in 0.
This happens because rightShift(int n, int b) is set to return 0 whenever b is greater than 30.
Actually the shift should result in 1, since we see the number as a normal bit sequence when logically shifting, as far as I can conclude from the implementation.
Probably this is not that a big problem, since I did not see this case happen in selfie.c. But it should be considered, for completeness.
If this holds, the same goes for leftShift(int n, int b). When shifting an odd number by 31 bits logically to the left, the result should be INT_MIN, not 0.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -1172,25 +1172,25 @@ int leftShift(int n, int b) {
 }
 
 int rightShift(int n, int b) {
-    // assert: b >= 0            
+    // assert: b >= 0
 
     if (n >= 0)
-        if (b > 30)
-            return 0;
-        else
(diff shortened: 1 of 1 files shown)


ckirsch:
Thanks a lot!

I thought of this case before but then forgot to handle it since I knew that I would never invoke it with b == 31.

I merged your pull request after polishing the code a bit more and fixing leftShift as well.

PR #9, opened 2016-08-19
Initialize registers and page table after allocating them

Blazefrost:


ckirsch:
I pulled this in but then polished the code a bit.

Issue #16, opened 2017-03-16
atoi() unreliable overflow handling


HugoPlatzer:
# Description of problem
the atoi() function does not always fail when an integer literal representing a number larger than 2^31 is processed.

# Example

`int main() {
    int a;
    a = 9999999999;
    return 0;
}`

compiles successfully while

`int main() {
    int a;
    a = 2222222222;
    return 0;
}`

doesn't.

# Cause

Multiplying , for example, 999999999 with 10 , as done at the end of the conversion for "9999999999" obviously causes an overflow. The resulting int value is however positive so the check `n < 0` does not trigger.

# Possible fix

Check for overflows in a different way: Store a value `x = INT_MAX  / 10 = 214748364` and a maximum digit `y = 8`. Before introducing a new digit c to the number n, fail if

1.  `n > x or (n == x and c > y)`, this is true if and only if `n * 10 + c` mathematically is larger than INT_MAX + 1, so the input string represents a number larger than abs(INT_MIN).

2. `n == INT_MIN`, since there are digits remaining, the input string represents too large a number.

See [this commit](https://github.com/cksystemsteaching/cc-2017-HugoPlatzer/commit/d1e494ee63a55e12d20970d6f763ad95500a54d1) for code (no pull request possible because of private repo).

ckirsch:
@HugoPlatzer Interesting! Can you do this in a way that `atoi` never produces an integer overflow, even when scanning INT-MIN? I'd like to remove all overflows from the code and this is a good opportunity. Feel free to fork `selfie` publicly and do your PR from there.

ckirsch:
@HugoPlatzer I just saw this again. Very nice! Without remembering your solution I came up with the same yesterday, see https://github.com/cksystemsteaching/selfie/pull/28

PR #18, opened 2017-03-18
Fixed signed integer overflow in atoi()

ChristianMoesl:
This is a possible solution for the issue #16. This issue consists of the problems:
1. Signed integer overflow, which is undefined behavior in C99 (problematic when compiling with gcc/clang, etc...)
2. The logic which should handle big decimal numbers

@julian.kanzler (on slack#cc2017) had a nice idea, to avoid integer multiplication, which is hard to check for an potential overflow. So i could come up with this solution.
All overflow checks rely on the following fact: n + c > INT_MAX   ->   n <= INT_MAX - c   (which is used to check for an possible overflow without the danger of an actual overflow).

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -106,8 +106,8 @@ int  stringLength(int* s);
 void stringReverse(int* s);
 int  stringCompare(int* s, int* t);
 
-int safeAddition(int n, int c);
-int safeMultiplication(int n, int c);
+int checkedAddition(int n, int c);
+int checkedMultiplication(int n, int c);
 
 int  atoi(int* s);
 int* itoa(int n, int* s, int b, int a, int p);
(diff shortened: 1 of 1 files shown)


ChristianMoesl:
Hmm, seams that the build failed due to a timeout of the travis build machine.

ckirsch:
@ChristianMoesl Let's talk about this next time we meet. Please remind me.

ChristianMoesl:
I totally forgot this pr.. . I'll remind you.

ckirsch:
@ChristianMoesl I just realized that I need this fixed for a paper I am writing. Could you please look into it as soon as you can? Please avoid introducing the `safe` procedures. Instead focus on line 1512 in `selfie.c` on the master branch. Check `n` against `214748363` for all `c` and `214748364` with `c < 8` before computing the next value of `n`. Return `INT_MIN` if `c == 8` and `-1` in all other cases. I hope this avoid any overflows. Otherwise, let me know.

ChristianMoesl:
@ckirsch i'm in holiday atm, so time is limited. But it should be possible to fix this, this week. 

_Sent from my Google Nexus 5X using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #25, opened 2017-05-07
true read, write, open and malloc syscalls

thuetter:


ckirsch:
@thuetter Is this PR still current or is it obsolete?

thuetter:
Obsolete in current version.

PR #28, opened 2017-08-02
Fixed integer overflows

ChristianMoesl:
This is a possible solution for the issue #16.

@ckirsch I used a slightly simplified version of your code and tested it carefully, so there should be no problems with it.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -1508,24 +1508,36 @@ int atoi(int* s) {
       // c was not a decimal digit
       return -1;
 
-    // assert: s contains a decimal number, that is, with base 10
-    n = n * 10 + c;
+    // assert: s contains a decimal number
+
+    // use base 10 but avoid integer overflow
+    if (n < INT_MAX / 10)
+      n = n * 10 + c;
+    else if (n == INT_MAX / 10) {
+      // In decimal representation of integers,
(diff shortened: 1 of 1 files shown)


selfie.c
```diff
@@ -1508,24 +1508,36 @@ int atoi(int* s) {
...
 
-    // assert: s contains a decimal number, that is, with base 10
-    n = n * 10 + c;
+    // assert: s contains a decimal number
+
+    // use base 10 but avoid integer overflow
+    if (n < INT_MAX / 10)
+      n = n * 10 + c;
+    else if (n == INT_MAX / 10) {
+      // In decimal representation of integers,
```
ckirsch: The numerical value of INT_MIN is always greater by one than of INT_MAX but not necessarily the least signnificant digit. Try to find a counterexample for some number of bits other than 32. Anyway, please remove the comment and resubmit the PR.

ChristianMoesl: I see. I wanted to describe, that the least significant digit in 2's kompliment (with base 10) of INT_MIN is indeed exactly 1 greater than the least significant digit of INT_MAX. This is not the case for a 2's kompliment representation with e.g. a base of 16. So we can only rely on this fact for decimal integer literals. 

Anyway, i'll remove the comment.


dez2erkompl

PR #31, opened 2017-09-02
Added support for compiling under Windows/Visual Studio 2017.

nubok:
Added support for compiling under Windows/Visual Studio 2017.

# How to compile

- Prerequisite: Visual Studio 2017 and its C/C++ compiler are installed

- Either start "x64_x86 Cross Tools Command Prompt for VS 2017" or "x86 Native Tools Command Prompt for VS 2017"

- 'cd' into selfie folder

- Type

`> nmake -f Makefile.win TARGETNAME`

to compile TARGETNAME. For compiling selfie.exe type

`> nmake -f Makefile.win selfie.exe`

(only the Makefile target "selfie" was renamed to "selfie.exe"; all other targets kept their existing name).

# For an explanation of Makefile.win:

"diff" was replaced by "fc" (cf. https://stackoverflow.com/q/6877238/497193). Unluckily fc has no "-q" option. Powershell seems to support 'diff', but "x64_x86 Cross Tools Command Prompt for VS 2017" and "x86 Native Tools Command Prompt for VS 2017" currently have no Powershell counterparts. So I considered using "fc" the better solution.

I get no trouble compilation problems when I removed the 'main(a,b)=main(a,char**argv)' definition - so I did remove it.

On the other hand I had to define "atoi" as something different (I chose "xatoi"), since otherwise one gets a name collission with the existing "atoi" function of the C Runtime

# What works and where are still bugs?

`> nmake -f Makefile.win sat`

seems to work.

For the tests in target "test":

`> selfie.exe -c selfie.c -o selfie1.m -s selfie1.s -m 2 -c selfie.c -o selfie2.m -s selfie2.s`

and

`> selfie.exe -c selfie.c -o selfie.m -m 2 -l selfie.m -m 1`

seem to work. On the other hand

`> selfie.exe -c selfie.c -o selfie3.m -s selfie3.s -y 3 -l selfie3.m -y 2 -l selfie3.m -y 2 -c selfie.c -o selfie4.m -s selfie4.s`

outputs:

```
selfie.exe: this is selfie compiling selfie.c with starc
selfie.exe: 191548 characters read in 7576 lines and 1032 comments
selfie.exe: with 105824(55.24%) characters in 31199 actual symbols
selfie.exe: 270 global variables, 311 procedures, 455 string literals
selfie.exe: 2094 calls, 809 assignments, 71 while, 627 if, 287 return
selfie.exe: 129704 bytes generated with 30755 instructions and 6684 bytes of data
[...]

PR #33, opened 2017-09-03
Additions to .gitignore

nubok:
# Additions to .gitignore that are helpful when working under Git on Windows

The Visual C++ compiler creates *.obj files (which have the same purpose as the *.o files under GNU/Linux). These are ignored. Also added a rule to ignore selfie.exe, too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 *.m
 *.s
+*.obj
 selfie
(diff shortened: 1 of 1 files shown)


ChristianMoesl:
@nubok This is also not needed as already stated in #31. But thanks for your interest.

PR #34, opened 2017-09-05
Fixed boot level zero detection

ChristianMoesl:
Fixed boot level zero detection by using a implementation detail of selfie's malloc.


ChristianMoesl:
This fix is needed by #37, so i will merge it into #37  and close this PR.

PR #35, opened 2017-09-05
Secure memory allocation

ChristianMoesl:
I noticed that it is never checked whether memory is correctly allocated. However, a correctly allocated memory is always required in order to continue the program. For this reason, i introduced `smalloc`. This procedures ensures correctly allocated memory or a safe program exit.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -1926,14 +1926,14 @@ int* smalloc(int size) {
   memory = malloc(size);
 
   if (size == 0)
-    // a null address is valid, if 0 bytes should be allocated
+    // any address including null
     return memory;
   else if ((int) memory == 0) {
     print(selfieName);
-    print((int*) ": could not allocate physical memory");
+    print((int*) ": malloc out of memory");
     println();
 
-    exit(EXITCODE_OUTOFPHYSICALMEMORY);
+    exit(EXITCODE_OUTOFVIRTUALMEMORY);
   }
 
   return memory;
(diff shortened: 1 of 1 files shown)


selfie.c
```diff
@@ -1917,6 +1918,27 @@ int roundUp(int n, int m) {
...
+  int* memory;
+
+  memory = malloc(size);
+
+  if (size == 0)
+    // a null address is valid, if 0 bytes should be allocated
+    return memory;
+  else if ((int) memory == 0) {
+    print(selfieName);
+    print((int*) ": could not allocate physical memory");
```
ckirsch: say something like `malloc out of memory`.

selfie.c
```diff
@@ -1917,6 +1918,27 @@ int roundUp(int n, int m) {
...
+    println();
+
+    exit(EXITCODE_OUTOFPHYSICALMEMORY);
```
ckirsch: Say `EXITCODE_OUTOFVIRTUALMEMORY`.

selfie.c
```diff
@@ -1917,6 +1918,27 @@ int roundUp(int n, int m) {
...
 
+int* smalloc(int size) {
+  // this procedure ensures a defined program exit,
+  // if no memory can be allocated
+  int* memory;
+
+  memory = malloc(size);
+
+  if (size == 0)
+    // a null address is valid, if 0 bytes should be allocated
```
ckirsch: any address including null

ChristianMoesl:
Merged this PR into [vipster branch](https://github.com/cksystemsteaching/selfie/tree/vipster).

PR #37, opened 2017-09-06
Increased virtual memory size to 4GB

ChristianMoesl:
Increased virtual memory size to 4GB.

## Prerequisites
To be able to allocated ~4GB of Memory, additional compiler flags are needed. On a Windows machine, the linker flag `-Wl,--large-address-aware` (gcc, clang) or `/LARGEADDRESSAWARE` (Visual C++) needs to be added. For Linux and Mac, i haven't tested it yet. I think the Windows Compiler flags won't work on this machines.

## Issues
For 4GB Address space we need unsigned arithmetics. Luckily Selfie has a lot of unsigned MIPS instructions (addu, subu, multu,..). Alltough the overflow detection might detect some overflows, which are in fact legal. For an example if the memory address  `0x7FFF FFFC` is increased by `4` we get `0x8000 0000` and an overflow warning.

Btw.: The instruction selfie DIVU implementation is more of a MIPS DIV instrcution. Signed division wouldn't work with the MIPS DIVU instrution.

## ToDo
We need to detect the operating system in the Makefile and add the corresponding linker flags for the os.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ test: selfie
 	diff -q selfie3.s selfie4.s
 	diff -q selfie1.m selfie3.m
 	diff -q selfie1.s selfie3.s
-	./selfie -c selfie.c -o selfie5.m -s selfie5.s -min 4 -l selfie5.m -y 2 -l selfie5.m -y 2 -c selfie.c -o selfie6.m -s selfie6.s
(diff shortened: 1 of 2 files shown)


ChristianMoesl:
Merged this PR into [vipster branch](https://github.com/cksystemsteaching/selfie/tree/vipster).

PR #43, opened 2017-10-18
Fixed main function signature in code examples for assignment 1

ChristianMoesl:
@cponcelets I noticed a little bug in the code examples, used in today's homework. It gives a warning during compilation with selfie. So basically it's not a big deal, but just to let you know.

### DIFF
--- a/manuscript/code/hello-world-minified.c
+++ b/manuscript/code/hello-world-minified.c
@@ -1,1 +1,1 @@
-uint64_t*foo;uint64_t*main(){foo="Hello World!    ";while(*foo!=0){write(1,foo,8);foo=foo+1;}}
+uint64_t*foo;uint64_t main(){foo="Hello World!    ";while(*foo!=0){write(1,foo,8);foo=foo+1;}}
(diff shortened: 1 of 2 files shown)


cponcelets:
I didn't noticed, thank you.

manuscript/code/hello-world-minified.c
```diff
@@ -1 +1 @@
-uint64_t*foo;uint64_t*main(){foo="Hello World!    ";while(*foo!=0){write(1,foo,8);foo=foo+1;}}
\ No newline at end of file
+uint64_t*foo;uint64_t main(){foo="Hello World!    ";while(*foo!=0){write(1,foo,8);foo=foo+1;}}
```
ckirsch: @ChristianMoesl The intention here is to have zero whitespace...even if there is a compiler warning...

ChristianMoesl: @ckirsch I see. But why do we have to have zero whitespaces? For Teaching? Isn't this a good example for the limitations of minification?

ckirsch: @ChristianMoesl To demonstrate that whitespace in C programs has no effect on semantics.

ChristianMoesl: @ckirsch Ok i see. I'm closing this.

PR #44, opened 2017-10-22
compile error flag

cponcelets:
Add a flag to terminate the compilation with an error.

- the syntax errors turn the flag on and continue the compilation
- selfie exit with compilation error if the flag is on


ckirsch:
@cponcelets this is more work. I can do this when I have time. It requires a bit more conceptualization. Error handling in compilers is more an art rather than engineering.

PR #47, opened 2017-12-02
Fix for potentially unmapped memory access

ChristianMoesl:
The page table caching algorithm touches  potentially unmapped memory, which lead's to an segmentation fault in some scenarios. The algorithm accesses a entry below the hiPage entry, which is not ensured to be already mapped.


ckirsch:
Interesting! Can you describe a scenario where this happens?

cponcelets:
I had this error running 4 forks in a while (I wanted to propose a PR too,
thank you Christian).

2017-12-02 19:34 GMT+01:00 Christoph Kirsch <notifications@github.com>:

ChristianMoesl:
@ckirsch for sure. I tried to visualize it:
image

ckirsch:
@ChristianMoesl Nice! This means there are two conditions to terminate caching: 1. the parent page table has not been touched at the position of `FrameForPage(table, page)`, or 2. there is a `0` stored there. However, why are you not checking that in the code right above. We could get the same error there, no? Also, we need a bit of commenting on the two conditions when to terminate caching. By the way, another fix would be to have `zalloc` zero memory also on boot levels other than zero, yet much slower, right?

ChristianMoesl:
I unified the solution for this problem. I used your recommendation and applied the same fix for the LO-ME interval in the page table caching algorithm.

PR #59, opened 2018-01-28
Added code length to binary format and improved ELF header

ChristianMoesl:


ckirsch:
@ChristianMoesl Looks like you also need to provide a bit more memory in some other calls in the makefile.

PR #60, opened 2018-03-26
Performance improvement for loading/storing variables

PhilippMayer92:
I reimplemented the code generation for loading/storing with offsets out of immediate range.
Therefore we used load_integer() with an additional ADD instruction to compute the base address and afterwards LD/SD, which is a total of 4 instructions. 
In my reimplementation we don't use load_integer() anymore but instead we use a single LUI and the additional ADD and LD/SD, which is a total of 3 instructions. The main point is, that we avoid the ADDI instruction from load_integer() and use the immediate part of the LD/SD instruction for that porpuse.
This saves us around thousand instructions for selfie.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -2892,23 +2892,36 @@ uint64_t* getVariableOrBigInt(uint64_t* variableOrBigInt, uint64_t class) {
 
 uint64_t load_variableOrBigInt(uint64_t* variableOrBigInt, uint64_t class) {
   uint64_t* entry;
+  uint64_t offset;
+  uint64_t lower;
+  uint64_t upper;
 
   // assert: n = allocatedTemporaries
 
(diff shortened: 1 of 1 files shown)


ckirsch:
Very nice!

You can still improve your code. It does the same thing in two places. Can you find the correct abstraction? Implement a procedure `load_upperBaseAddress` that works for both LD and SD operations.  It's ok to compute `lower` twice. Avoid the word "suitable" in comments. It's too vague. Avoid `return getType` twice.

ChristianMoesl:
How about if we introduce some constants for e.g. "12" and "20"? For an example something like IMMEDIATE_BITDWIDTH.

PhilippMayer92:
Hmm...I'm not completelly convinced of that idea, because then we should also use constants for the length of register addresses, the length of opcodes and so on...

PR #61, opened 2018-03-29
overflow checking for add and sub w-intervals

cponcelets:
Adjust the **combinedCardinality** function.
Check abstract overflows is done with **isWOverflow**, this function returns 1 or 0 if an overflow is occurring or not according to the two w-intervals in input.
This is done without **cardinality** (only *up-low*) preventing from the wrap of *+1*, thus **cardinality** is removed.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -1101,8 +1101,7 @@ void initSymbolicEngine();
 
 void printSymbolicMemory(uint64_t svc);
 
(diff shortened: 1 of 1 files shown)


ckirsch:
@cponcelets Since I lost track of what exactly we need to do to make the engine become complete I am holding off on this PR.

PR #62, opened 2018-03-31
Fixed exit code sign extension

ChristianMoesl:
Fixed exit code sign extension as discussed with @cponcelets .


ckirsch:
@ChristianMoesl Thank you for cleaning this up. However, please go a step further and check how to use exit codes in C. Apparently, they are meant to be in the range 0-255 with 0 indicating no errors. I suggest you remove the whole sign code and just use positive error codes. Also, there are a few lines missing in your PR that need work, e.g. in read and write, search for signShrink.

ChristianMoesl:
@ckirsch I tested the return value range with a simple C program, where I just return a specific value in the main function. Turned out, that we get up to 32 bit signed exit codes in power shell on windows. So I think we should stick with that. 

I did look into read etc. The problem here is, that the real signature of read is “int  read(...)” therefore we have to stick with 32bit integers in this place. Therefore we have to shrink negative numbers.

And yeah, I also think that positive exit codes are more appropriate. I will replace the negative exit codes with positive ones.

ckirsch:
@ChristianMoesl I am not sure I understand. If we only use positive error codes we don't need any shrinking, right? But should be ok with that even for Windows, no?

ChristianMoesl:
@ckirsch I merged your changes into my PR. I will try to explain my point again. 

`signShrink` in `implementExit`:
For this case you are indeed right. If we don't use negative exit codes in selfie, we don't need that one `signShrink` call in `implementExit`. Anyhow, i don't think we should do that, because we should also consider the case, where we compile some other c code with selfie. 

`signShrink` in `implementRead`, `implementWrite` and `implementOpen`:
In this case, we can not get rid of the `signShrink` call, because these syscalls all return `int` and have to return `-1` in some cases.

PR #63, opened 2018-04-18
semantics.md

PhilippMayer92:


ckirsch:
@PhilippMayer92 Thank you! I actually got inspired by this and fixed the precedence of the unary - operator. So, we should be equivalent to C on that level now. Let me suggest that you remove that part and then focus on two classes of difference: 1. integer literals in C* are unsigned 64-bit instead of signed 32-bit in C and 2. strings in C* are arrays of unsigned 64-bit integers rather than of 8-bit characters. Then list examples of code where this becomes apparent. Is that all? I am not sure?

PhilippMayer92:
@ckirsch  Do you have an example for me with semantical differences for strings?
I'm also not sure whether this is all. But I think, these are all known differences.

ckirsch:
@PhilippMayer92 At least the distinction of integer and string literals is a good start. String literals are handled differently in C* because they are of type `uint64_t*` rather than `char*`. This means accessing strings will behave differently. For example, consider `*"Hello World!"`. In C this is `72` (ASCII for H) whereas in C* this is `8022916924116329800` (ASCII for the first 8 characters).

PR #68, opened 2018-05-20
Multi line comment support

MarioStrohmeier:
Added support for multi line comments with "/*" as start and "*/" as end markers.
Also changed the first comment block to a multi line comment, to show that it works.


selfie.c
```diff
@@ -2291,27 +2293,62 @@ uint64_t isCharacterWhitespace() {
...
         return character;
       else
         // count the characters in comments as ignored characters
         // line feed and carriage return are counted below
         numberOfIgnoredCharacters = numberOfIgnoredCharacters + 1;
 
+    } else if (multiLineComment) {
+      getCharacter();
+
+      // comments end with */
```
ckirsch: // multi line comments end ...

selfie.c
```diff
@@ -2291,27 +2293,62 @@ uint64_t isCharacterWhitespace() {
 }
 
 uint64_t findNextCharacter() {
-  uint64_t inComment;
+  uint64_t singleLineComment;
```
ckirsch: Please rename to `inSingleLineComment`.

selfie.c
```diff
@@ -2291,27 +2293,62 @@ uint64_t isCharacterWhitespace() {
...
+  uint64_t multiLineComment;
```
ckirsch: Same here, `inMultiLineComment`.

selfie.c
```diff
@@ -2291,27 +2293,62 @@ uint64_t isCharacterWhitespace() {
...
         // count the characters in comments as ignored characters
         // line feed and carriage return are counted below
         numberOfIgnoredCharacters = numberOfIgnoredCharacters + 1;
 
+    } else if (multiLineComment) {
+      getCharacter();
+
+      // comments end with */
+      if (character == CHAR_ASTERISK) {
+      	getCharacter();
```
ckirsch: Indentation is 2 spaces, not 1.

selfie.c
```diff
@@ -2291,27 +2293,62 @@ uint64_t isCharacterWhitespace() {
...
+
+      	if (character == CHAR_SLASH) {
+      	  multiLineComment = 0;
+         
+         // count */ as two ignored characters
+         numberOfIgnoredCharacters = numberOfIgnoredCharacters + 2;
+
+      	  getCharacter(); // next character after */
+      	} else if (character == CHAR_EOF)
```
ckirsch: Can you try to avoid duplicating the code here? Check for whitespace before checking for `inMultiLineComment`.

selfie.c
```diff
@@ -2286,25 +2288,32 @@ uint64_t isCharacterWhitespace() {
...
+    return 1;
   else
     return isCharacterNewLine();
 }
 
 uint64_t findNextCharacter() {
-  uint64_t inComment;
+  uint64_t inSingleLineComment;
+  uint64_t inMultiLineComment;
+  uint64_t readWhitespaceInMultiLineComment;
```
ckirsch: Can you avoid using such a flag? How about first checking `inSingleLineComment`, then `inMultiLineComment`, then whitespace, then slash?

selfie.c
```diff
@@ -2286,25 +2288,32 @@ uint64_t isCharacterWhitespace() {
     return 1;
   else if (character == CHAR_TAB)
     return 1;
+  else if (character == CHAR_EOF)
```
ckirsch: Please avoid changing this function. It is also used in the DIMACS parser.

selfie.c
MarioStrohmeier: That would require to check for line feeds and EOF in `inMultiLineComment` again, since both read characters in `inMultiLineComment` could be those two and `isCharacterWhitespace()` is never reached. And in turn the code duplication would be back.

ckirsch: Interesting. I am wondering if there is an abstraction that would simplify the code and avoid code duplication.

MarioStrohmeier: The `lineNumber` could be incremented in `getCharacter()`, that would mean no increments anywhere else and would only leave EOF.

ckirsch:
@MarioStrohmeier Please close the PR and then open a new PR. My recent changes make that necessary.

PR #69, opened 2018-05-24
Added defines

leodennis:
Added the simplest form of defines as they are in the C standard.

This makes it possible to define NULL which makes the code much more readable.
Also you can define variables like NUMBEROFREGISTERS so they cannot be changed.

If you want you can also define boolean states with true as 1 and false as 0.

Currently it is saved in the libraty_symbol_table for performace reasons, other possible solutions are a define_symbol_table or a hash table for the global_symbol_table.

I'm also not too sure if the grammar.md is changed correctly.

As adding defines is part of a precompilation the compile_define does not follow the normal rules of scanner and parser.
Also this merge adds infrastructure for other precompilation possibilities if they are wished.

TODO: Error checking is currently not implemented (identifiers used by defines cannot be used by functions or other variables)

**EDIT:**
OK rigth now travis failes, don't know jet why because it works with the makefile of the compiler class...


ckirsch:
This is a nice idea but the code requires a lot of work. I am happy to help but that would need to be done in person.

ckirsch:
@leodennis Same here. Please close the PR and then open a new PR based on the current master branch.

PR #71, opened 2018-05-25
Added defines

leodennis:
Added the simplest form of defines as they are in the C standard.

This makes it possible to define NULL which makes the code much more readable.
Also you can define variables like NUMBEROFREGISTERS so they cannot be changed.

If you want you can also define boolean states with true as 1 and false as 0.

I also changed to global symbol table into a simple fixed size hash table.

I'm also not too sure if the grammar.md is changed correctly.

As adding defines is part of a precompilation the compile_define does not follow the normal rules of scanner and parser.
Also this merge adds infrastructure for other precompilation possibilities if they are wished.

**NEW EDIT:**
I will update to the new master branch when I have time...


ckirsch:
@leodennis @ChristianMoesl I close this PR but keep it in mind for future reference.

PR #73, opened 2018-05-29
Malloc semantics

ChristianMoesl:
Described malloc semantic differences of C and C\*


ChristianMoesl:
oh god. I got something wrong here. I will fix this

PR #74, opened 2018-06-07
Added printf

leodennis:
This pull request adds a prinf function and support for variable argument lists.

I haven't changed all the prints but I will if this is actually integrated into the selfie master branch...
I only changed some prints for a demonstration and proof of work.

As it seems impossible to access the parameters when compiled with gcc and clang the variable arguments of printf are converted into an array (using a define passed to the compiler) and the printf_print function is called insted.
The only ugly part is that the printf function has to be sourrounded by parenthesis so it won't be replaced by the define.

Also it adds checking if the number of arguments are correct at compile_call and it allows getting an address of a variabe using &

Also don't get `signExtend(getBits(offset, 0, 12), 12)` at `load_variableOrBigInt` so I only implemented variable arguemnts for `if (isSignedInteger(offset, 12)) {`


ckirsch:
@leodennis @ChristianMoesl Same here!

PR #75, opened 2018-06-11
added global hash table

leodennis:
Changed the global symbol table to a hash table.

It is a simple fixed size symbol table which saves entries in a linked list at collision.

I think it will speed up the compilation of selfie.
This is expecially useful in the systems and compiler class because you do `make clean test` quite a lot...

I think it speeds up self compilation up to 40% and around 66% of the hast table is filled with entries.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -468,6 +468,13 @@ void setValue(uint64_t* entry, uint64_t value)        { *(entry + 5) = value; }
 void setAddress(uint64_t* entry, uint64_t address)    { *(entry + 6) = address; }
 void setScope(uint64_t* entry, uint64_t scope)        { *(entry + 7) = scope; }
 
+// --------------------------- HASH TABLE --------------------------
+
+uint64_t GLOBAL_HASH_TABLE_SIZE = 1024;
+
+uint64_t getHashValue(uint64_t* key);
+uint64_t* getHashEntry(uint64_t* hash_table, uint64_t* key);
+
 // ------------------------ GLOBAL CONSTANTS -----------------------
 
 // classes
@@ -489,7 +496,7 @@ uint64_t LIBRARY_TABLE = 3;
 // ------------------------ GLOBAL VARIABLES -----------------------
 
 // table pointers
-uint64_t* global_symbol_table  = (uint64_t*) 0;
+uint64_t* global_hash_symbol_table = (uint64_t*) 0;
 uint64_t* local_symbol_table   = (uint64_t*) 0;
 uint64_t* library_symbol_table = (uint64_t*) 0;
 
@@ -500,13 +507,23 @@ uint64_t numberOfStrings         = 0;
 // ------------------------- INITIALIZATION ------------------------
 
 void resetSymbolTables() {
-  global_symbol_table  = (uint64_t*) 0;
+  uint64_t i;
+  
+  if (global_hash_symbol_table == (uint64_t*) 0)
+    global_hash_symbol_table = smalloc(GLOBAL_HASH_TABLE_SIZE * SIZEOFUINT64STAR);
   local_symbol_table   = (uint64_t*) 0;
   library_symbol_table = (uint64_t*) 0;
 
   numberOfGlobalVariables = 0;
   numberOfProcedures      = 0;
   numberOfStrings         = 0;
+  
+  // zero hash table
+  i = 0;
+  while (i < GLOBAL_HASH_TABLE_SIZE) {
+    *(global_hash_symbol_table + i) = 0;
+    i = i + 1;
+  }
 }
 
 // -----------------------------------------------------------------
@@ -590,6 +607,8 @@ void resetParser() {
 // -----------------------------------------------------------------
 // ---------------------- MACHINE CODE LIBRARY ---------------------
 // -----------------------------------------------------------------
+
+uint64_t gp; // has to be initialized in emitStart
 
 void emitLeftShiftBy(uint64_t reg, uint64_t b);
 void emitProgramEntry();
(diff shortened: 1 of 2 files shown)


selfie.c
```diff
@@ -2736,8 +2755,10 @@ void createSymbolTableEntry(uint64_t whichTable, uint64_t* string, uint64_t line
   // create entry at head of symbol table
   if (whichTable == GLOBAL_TABLE) {
     setScope(newEntry, REG_GP);
-    setNextEntry(newEntry, global_symbol_table);
-    global_symbol_table = newEntry;
+    
+    // add to hash table
+    setNextEntry(newEntry, getHashEntry(global_hash_symbol_table, string));
+    *(global_hash_symbol_table + getHashValue(string)) = (uint64_t) newEntry;
```
ChristianMoesl: You are introducing a new bug here. As you might know, the essence of an hash function is, to map a value of arbitrary size to a smaller value. So as your set of keys increase, also the chance of equal hash values will increase. In this example this would result in overridden symbol table entries. I think closed hashing is the solution to go.

selfie.c
```diff
@@ -468,6 +468,13 @@ void setValue(uint64_t* entry, uint64_t value)        { *(entry + 5) = value; }
 void setAddress(uint64_t* entry, uint64_t address)    { *(entry + 6) = address; }
 void setScope(uint64_t* entry, uint64_t scope)        { *(entry + 7) = scope; }
 
+// --------------------------- HASH TABLE --------------------------
+
+uint64_t GLOBAL_HASH_TABLE_SIZE = 1024;
```
ChristianMoesl: I'm not completely sure, if exchanging a dynamic data structure with a static structure is good in this case, especially if you don't handle errors. Possible solutions would be
-  Grow your hash table size, as you need it (maybe with a log(n) complexity in the number of entries)
-  Detect when the symbol table is full or reaches a specific load factor (because performance suffers from an high load factor) and exit with an error message

ckirsch:
@ChristianMoesl @leodennis I think having a simple hash table in selfie is in principle a good idea for two reasons. 1. performance apparently improves significantly which makes the experience of using selfie in class better, and 2. the educational value of having a hash table, in addition to lists and "structs". Now, to make this happen we need a super simple and clean implementation. I would not mind to have a static size on the hash table and even a hash function that may not scale beyond selfie itself. The rule should be that it works well for selfie but not necessarily for other code.

ChristianMoesl:
Okay, sounds good. 

For the hash, you could use the guarantee that the first double word in a string (the key) is always allowed to be read. So the simplest hash function could be something like: 
```c
uint64_t hash(uint64_t* key) { 
  return *key % HASH_TABLE_SIZE; 
}
```
But because there are many functions starting with the same three chars (e.g. get, set), it might be not optimal. Another approach could be this:
```c
uint64_t hash(uint64_t* key) { 
  return (*key + getHighWord(*key)) % HASH_TABLE_SIZE; 
}
```
But the best solution I think is, to use the guarantee of unique heap pointer addresses in selfie and use this value for the hash:
```c
uint64_t hash(uint64_t* key) { 
  return ((uint64_t)key) / SIZEOFUINT64 % HASH_TABLE_SIZE; 
}
```

ckirsch:
Indeed! Nice! Using `leftShift` rather than `/` :)

ckirsch:
`rightShift`, I mean :)

leodennis:
@ChristianMoesl The hash table works perfectly fine without any bugs, when two keys result in the same hash value they are not overwritten but saved as a linked list just like the original global_symbol_table was.
The the better the hash function is the better the perfomance gets because it won't need to search the linked list at collision.

ckirsch:
@leodennis Sure. Please incorporate the heap pointer hash function by @ChristianMoesl and my other comments. I will then pull your code in. Thanks!

leodennis:
@ChristianMoesl I tired out your hash functions.
My hash functions fills the hash_table around 66%
Your first hash function fills the table around 13%
Your second hash function fills the table around 35%
And your last function which I was meant to implement (@ckirsch ) does not work because the address of a global variable may change which will result in a different hash...

ckirsch:
@Leo
Try the address of a string in memory, not the address of the global variable containing the address of the string.

leodennis:
@ckirsch 
To do that I would need to search the symbol table for the string to get the address in memory. Which obviously is not an option.
Or is there something I am missing?

ChristianMoesl:
@leodennis Okay I see now my confusion with the hash table. You use separate chaining for your hash table. That is indeed the best fitting approach with minimal modification. But I would argue, that from an engineering point of view, this isn't realy clear when reading your code. We should give clear meaning for that, when we insert/search for entries in a hash table, which is in this approach a hash table with linked lists as entries. 

And yeah, my last hash function won't work in that way, I thought about it. The guarantee of unique addresses is an advantage and the point which makes this not work at all, because strings with equals content will get different addresses.

ChristianMoesl:
How about something like this?
```c
uint64_t hash(uint64_t* key) { 
  return (*key + rightShift(*key, 16) + rightShift(*key, 32) + rightShift(*key, 48)) % HASH_TABLE_SIZE; 
}
```
If we are able to be nearly as good as your initial hash function, but faster, it would be good I think. We should really use the power of 64bit load instructions.

ckirsch:
@leodennis @ChristianMoesl True! Addresses won't work. Yes, a hash over the first machine word may still be the best way to do this. Ultimately, the best such hash functions depends on the distribution of the first 8 characters in all identifiers as well as all string and big integer literals.

leodennis:
@ChristianMoesl 
Your last hash function fills the hash table 53% (and with a hash table size of 512 73%).
I guess that is already quite good.
The only problem is that when we do `rightShift(*key, 32)` and `rightShift(*key, 48)` (shifting *key with any number greater that 31) selfie will print out the statistics and the abort compiling without giving me any error message...

ChristianMoesl:
@leodennis This are already good results. We could also compare it to a function, which adds all 8 bytes of a word. It ultimately comes down to how big is the hash table. Each addition should be 

 I think the hash table shouldn't be too small, because memory is not that of a problem as performance in selfie. 

In the end we have to do some performance testing as well. Some code refinements (more general hash table) are also needed to get this PR into the master branch. 

About the problem with `rightShift`: This is strange. Are you testing this on the current master branch?

ckirsch:
@leodennis @ChristianMoesl Thanks a lot for your initiative and feedback! I went ahead and implemented a simple version of a hash table and pushed it into the master branch. According to my measurements symbols are now found after no more than two iterations on average through any list of symbols, down by two orders of magnitude when not using the hash table. Makes the build significantly faster! Very cool.

PR #77, opened 2018-06-22
Spike: Improved string functions to work with bytewise aligned strings

ChristianMoesl:
Spike gives us only byte wise aligned strings on program start, such that all strings in the `argv` table are not double word aligned. 

My initial approach was, to align the whole table in the main function by copying it in the right way. This worked, but wasn't that great, because every other program than selfie would had to do that also in the main function.

So I had to do that before the main function get's called, which obviously means that this would had to be done in assembler. That's not an option because of all the complexity it introduces.

My final approach, which is this PR is, to just allow byte wise aligned strings in selfie. And because every string function uses `loadCharacter()` or `storeCharacter()`, it's easy to do this.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -123,6 +123,9 @@ uint64_t signShrink(uint64_t n, uint64_t b);
 uint64_t  loadCharacter(uint64_t* s, uint64_t i);
 uint64_t* storeCharacter(uint64_t* s, uint64_t i, uint64_t c);
 
+uint64_t  stringAddressOffset(uint64_t* s);
+uint64_t* stringAddressAligned(uint64_t* s);
+
 uint64_t stringLength(uint64_t* s);
 void     stringReverse(uint64_t* s);
 uint64_t stringCompare(uint64_t* s, uint64_t* t);
@@ -1763,6 +1766,10 @@ uint64_t loadCharacter(uint64_t* s, uint64_t i) {
   // assert: i >= 0
   uint64_t a;
 
+  // align string address for double word access
+  i = i + stringAddressOffset(s);
+  s = stringAddressAligned(s);
+
   // a is the index of the double word where
   // the to-be-loaded i-th character in s is
   a = i / SIZEOFUINT64;
@@ -1775,6 +1782,10 @@ uint64_t* storeCharacter(uint64_t* s, uint64_t i, uint64_t c) {
   // assert: i >= 0, 0 <= c < 2^8 (all characters are 8-bit)
   uint64_t a;
 
+  // align string address for double word access
+  i = i + stringAddressOffset(s);
+  s = stringAddressAligned(s);
+
   // a is the index of the double word where
   // the with c to-be-overwritten i-th character in s is
   a = i / SIZEOFUINT64;
@@ -1784,6 +1795,16 @@ uint64_t* storeCharacter(uint64_t* s, uint64_t i, uint64_t c) {
   *(s + a) = (*(s + a) - leftShift(loadCharacter(s, i), (i % SIZEOFUINT64) * 8)) + leftShift(c, (i % SIZEOFUINT64) * 8);
 
   return s;
+}
+
+uint64_t stringAddressOffset(uint64_t* s) {
+  // offset of a bytewise address to a double word aligned address
+  return ((uint64_t)s) % SIZEOFUINT64;
+}
+
+uint64_t* stringAddressAligned(uint64_t* s) {
+  // double word aligned address of a bytewise aligned address
+  return (uint64_t*) (((uint64_t)s) - stringAddressOffset(s));
 }
 
 uint64_t stringLength(uint64_t* s) {
@@ -8716,7 +8737,7 @@ uint64_t up_loadString(uint64_t* context, uint64_t* s, uint64_t SP) {
   uint64_t bytes;
   uint64_t i;
 
-  bytes = roundUp(stringLength(s) + 1, REGISTERSIZE);
+  bytes = roundUp(stringAddressOffset(s) + stringLength(s) + 1, REGISTERSIZE);
 
   // allocate memory for storing string
   SP = SP - bytes;


ChristianMoesl:
@ckirsch Christian Barthel didn't mentioned this problem in his thesis and i can't see any fix for that in the code. So the problem clearly arised either by using an updated Spike emulator or by the switch to 64bit.

selfie.c
```diff
@@ -1763,6 +1766,10 @@ uint64_t loadCharacter(uint64_t* s, uint64_t i) {
   // assert: i >= 0
   uint64_t a;
 
+  // align string address for double word access
```
ckirsch: double-word access

selfie.c
```diff
@@ -1775,6 +1782,10 @@ uint64_t* storeCharacter(uint64_t* s, uint64_t i, uint64_t c) {
   // assert: i >= 0, 0 <= c < 2^8 (all characters are 8-bit)
   uint64_t a;
 
+  // align string address for double word access
```
ckirsch: double-word access

selfie.c
```diff
@@ -1786,6 +1797,16 @@ uint64_t* storeCharacter(uint64_t* s, uint64_t i, uint64_t c) {
   return s;
 }
 
+uint64_t stringAddressOffset(uint64_t* s) {
+  // offset of a bytewise address to a double word aligned address
```
ckirsch: double-word-aligned address

selfie.c
```diff
@@ -1786,6 +1797,16 @@ uint64_t* storeCharacter(uint64_t* s, uint64_t i, uint64_t c) {
...
+  return ((uint64_t)s) % SIZEOFUINT64;
+}
+
+uint64_t* stringAddressAligned(uint64_t* s) {
+  // double word aligned address of a bytewise aligned address
```
ckirsch: double-word-aligned address ... bytewise address

selfie.c
```diff
@@ -8716,7 +8737,7 @@ uint64_t up_loadString(uint64_t* context, uint64_t* s, uint64_t SP) {
   uint64_t bytes;
   uint64_t i;
 
-  bytes = roundUp(stringLength(s) + 1, REGISTERSIZE);
+  bytes = roundUp(stringAddressOffset(s) + stringLength(s) + 1, REGISTERSIZE);
```
ckirsch: I don't understand this line. The address of the uploaded string will be double-word-aligned, right?

ChristianMoesl: The addresses of the uploaded strings will not be double-word-aligned. Only raw memory is copied here. So if the strings are not aligned on the host, they will also not be aligned in the hosted machine.

selfie.c
ChristianMoesl: a5f226800f143a1e3b9ff252104ad53883e9c029

selfie.c
ChristianMoesl: a5f226800f143a1e3b9ff252104ad53883e9c029

selfie.c
ChristianMoesl: a5f226800f143a1e3b9ff252104ad53883e9c029

selfie.c
ChristianMoesl: a5f226800f143a1e3b9ff252104ad53883e9c029

selfie.c
ckirsch: Ok, but in your code you do not store the actual non-double-word aligned beginning of a string in the `argv` array but the address of the surrounding double word that contains the beginning of the string. I believe the confusion is because the value of `*s` is unclear if `s` is not double-word-aligned. What is the value?

ChristianMoesl: Okay, that's actually really interesting. My code is wrong, as you stated. I wanted to upload the strings unaligned to the emulator. But then in the argv table there will be aligned strings don't matching the alignment of the strings before. So why does this work at all? 

At least Spike allows the `ld` instruction to load from all bytewise adresses. Therefore we implicitly align the strings in `up_loadString`, when accessing `s` (which is a bytewise address as you correctly said). 

This behaviour of Spike is very surprising for me, but interestingly the riscv documentation for `ld` says:   
> The execution environment will define what portions of the address space are legal to access.

I think in order for teaching, we should stick with current behaviour to allow `ld` access only on double-word-aligned addresses. If you agree with me, I will fix this problem by storing the correct bytewise string addresses in `argv` and fix the alignment of `s`.

ckirsch: Here's what I'd try to do: `up_loadString` should take the potentially unaligned string and upload it to the virtual space into aligned machine words. For this to work you need to figure out what `*s` on an unaligned address actually returns (on boot level zero). How does that sound?

The other thing is to find out why your current version actually works. This could also be valuable to know...can you try?

ChristianMoesl: That's actually what it does right now. `s` is a bytewise address, which is read like you would expect it. I did an little test for that. I started selfie with spike with the argument `'abcdefghijklmnop'` and printed an memory dump, where I increase the address by 1 in every iteration. This gives me:
```
0x6867666564636261[@0x7F7E9BE6]
0x6968676665646362[@0x7F7E9BE7]
0x6A69686766656463[@0x7F7E9BE8]
0x6B6A696867666564[@0x7F7E9BE9]
0x6C6B6A6968676665[@0x7F7E9BEA]
0x6D6C6B6A69686766[@0x7F7E9BEB]
0x6E6D6C6B6A696867[@0x7F7E9BEC]
0x6F6E6D6C6B6A6968[@0x7F7E9BED]
0x706F6E6D6C6B6A69[@0x7F7E9BEE]
```

in ASCII:
```
hgfedcba
ihgfedcb
jihgfedc
kjihgfed
lkjihgfe
mlkjihgf
nmlkjihg
onmlkjih
ponmlkji
```

So to conclude. If we want to allow these double-word accesses on bytewise addresses on spike boot level, this PR is really worthless. I tested it now without this patch with spike. And yeah it works out of to box. I think we are done here 😅

ckirsch:
Interesting! I learned something here. Ok, feel free to close this one.

PR #78, opened 2018-06-22
Spike: introduced brk and implemented malloc in assembler

ChristianMoesl:
**Problem**
In order to be compatible with Spike, we have to adapt to available memory management possibilities of this system. One of my criteria was, to be compatible with Spike out-of-the-box, without any changes to Spike. This brings me to the available syscall`s in Spike. There is an old fashioned `brk`(no `sbrk`) and `mmap`. 

**mmap**
This was of course the way to go, I thought. Turned out, that it's not that easy to use this syscall and to be compatible with multiple OS when bootstrapping. For an example, I tried to use `mmap` on MacOS and on Linux to allocate memory space. Turned out, that they need different parameters depending on the host machine. (MacOS: MAP_ANON (deprecated), Linux: MAP_ANONYMOUS) The second problem with mmap was, that I just wasn't't able to use it in Spike. I got errors and errors. Maybe they just don't support it, to allocate memory in this way.

**brk**
Good old `brk` to the rescue. It's anyway easier to implement, I thought. After a little bit of research I realised, that there are two semantically different versions of `brk`. The `C` version returns `-1` or `0` whereas the syscall returns either the the previous brk value or the new brk value, depending on if there was an error or not. That results obviously in an bootstrapping problem for selfie. 

I fixed this with an nasty hack (it really is). I implemented `brk` as syscall in the emulator and malloc as an library function in assembler of the compiler, such that selfie generated binaries will always have malloc as portable library function, which calls the syscall version of `brk` internally to allocate memory. For that, I had to introduce a convention for storing the current bump pointer in assembler. I used an unused Calle saved  register (`s1`) for that, like we use for the frame pointer (`s0`).

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -910,22 +910,20 @@ uint64_t down_loadString(uint64_t* table, uint64_t vstring, uint64_t* s);
 void     implementOpen(uint64_t* context);
 
 void emitMalloc();
-void implementMalloc(uint64_t* context);
+void implementBrk(uint64_t* context);
 
 // ------------------------ GLOBAL CONSTANTS -----------------------
 
-uint64_t debug_read   = 0;
-uint64_t debug_write  = 0;
-uint64_t debug_open   = 0;
-uint64_t debug_malloc = 0;
+uint64_t debug_read  = 0;
+uint64_t debug_write = 0;
+uint64_t debug_open  = 0;
+uint64_t debug_brk   = 0;
 
 uint64_t SYSCALL_EXIT  = 93;
 uint64_t SYSCALL_READ  = 63;
 uint64_t SYSCALL_WRITE = 64;
 uint64_t SYSCALL_OPEN  = 1024;
-
-// TODO: fix this syscall for spike
-uint64_t SYSCALL_MALLOC = 222;
+uint64_t SYSCALL_BRK   = 214;
 
 // -----------------------------------------------------------------
 // ----------------------- HYPSTER SYSCALLS ------------------------
@@ -4519,6 +4517,29 @@ void emitStart() {
   // fixup jump at address 0 to here
   fixup_relative_JFormat(0, binaryLength);
 
+  // initialize s1 with the current bump pointer
+  emitADDI(REG_A0, REG_ZR, 0);
+  emitADDI(REG_A7, REG_ZR, SYSCALL_BRK);
+
+  emitECALL();
+
+  talloc();
+
+  // align the current bump pointer for double word access
+  emitADDI(REG_A0, REG_A0, SIZEOFUINT64 - 1);
+  emitADDI(currentTemporary(), REG_ZR, SIZEOFUINT64);
+  emitREMU(currentTemporary(), REG_A0, currentTemporary());
+  emitSUB(REG_A0, REG_A0, currentTemporary());
+
+  tfree(1);
+
+  // set program brk to the aligned bump pointer
+  emitADDI(REG_A7, REG_ZR, SYSCALL_BRK);
+
+  emitECALL();
+
+  emitADDI(REG_S1, REG_A0, 0);
+
   // calculate the global pointer value accommodating 6 more instructions
   gp = ELF_ENTRY_POINT + binaryLength + 6 * INSTRUCTIONSIZE + allocatedMemory;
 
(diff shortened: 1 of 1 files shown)


ChristianMoesl:
@ckirsch Christian Barthel had similar problems as I with `mmap`. His solution to make memory management work, was to patch the `sbrk` syscall into **pk** and use `sbrk` in the `malloc` asm library function.

selfie.c
```diff
@@ -4519,6 +4517,29 @@ void emitStart() {
...
 
+  // initialize s1 with the current bump pointer
+  emitADDI(REG_A0, REG_ZR, 0);
+  emitADDI(REG_A7, REG_ZR, SYSCALL_BRK);
+
+  emitECALL();
+
+  talloc();
+
+  // align the current bump pointer for double word access
```
ckirsch: double-word access

selfie.c
```diff
@@ -4519,6 +4517,29 @@ void emitStart() {
...
+  emitADDI(REG_A0, REG_A0, SIZEOFUINT64 - 1);
```
ckirsch: This block repeats below. Can you write a procedure `emitAlign` or something similar to `emitLeftShiftBy`?

selfie.c
```diff
@@ -4519,6 +4517,29 @@ void emitStart() {
...
+  emitSUB(REG_A0, REG_A0, currentTemporary());
+
+  tfree(1);
+
+  // set program brk to the aligned bump pointer
+  emitADDI(REG_A7, REG_ZR, SYSCALL_BRK);
+
+  emitECALL();
+
+  emitADDI(REG_S1, REG_A0, 0);
```
ckirsch: Add a comment

selfie.c
```diff
@@ -5948,34 +5969,74 @@ void emitMalloc() {
   // assuming that page frames are zeroed on boot level zero
   createSymbolTableEntry(LIBRARY_TABLE, (uint64_t*) "zalloc", 0, PROCEDURE, UINT64STAR_T, 0, binaryLength);
 
-  emitLD(REG_A0, REG_SP, 0); // size
+  talloc();
```
ckirsch: Add a comment

ckirsch: Also, only `talloc` as late as possible right before use.

selfie.c
ckirsch: Also, add a comment with the formula that is computed here.

selfie.c
```diff
@@ -5948,34 +5969,74 @@ void emitMalloc() {
...
+  emitSUB(currentTemporary(), currentTemporary(), previousTemporary());
+
+  // call brk syscall to set the current bump pointer
+  emitADD(REG_A0, currentTemporary(), REG_S1);
+  emitADDI(REG_A7, REG_ZR, SYSCALL_BRK);
 
   emitECALL();
 
-  emitJALR(REG_ZR, REG_RA, 0);
+  // return 0 iff memory allocation failed
```
ckirsch: Try to add a comment with pseudo C* code that abstracts from the assembly, or at least some more explanation as to what is going on here.

selfie.c
```diff
@@ -5948,34 +5969,74 @@ void emitMalloc() {
...
+
+  emitJALR(REG_ZR, REG_RA,0);
 }
 
-void implementMalloc(uint64_t* context) {
+void implementBrk(uint64_t* context) {
   // parameter
+  uint64_t bump;
+
+  uint64_t previousBump;
```
ckirsch: This should be `// local variables`, see `implementOpen`, for example.

selfie.c
```diff
@@ -5948,34 +5969,74 @@ void emitMalloc() {
...
   uint64_t size;
+  uint64_t valid;
 
-  // local variable
-  uint64_t bump;
+  previousBump = getBumpPointer(context);
```
ckirsch: Do this later right before it is required.

selfie.c
```diff
@@ -5948,34 +5969,74 @@ void emitMalloc() {
...
 
-  size = *(getRegs(context) + REG_A0);
+  bump = *(getRegs(context) + REG_A0);
+
+  if (debug_brk) {
+    print(selfieName);
```
ckirsch: Use `printf2`.

selfie.c
```diff
@@ -5948,34 +5969,74 @@ void emitMalloc() {
...
-  bump = getBumpPointer(context);
+  if (bump >= previousBump)
+    if (bump < *(getRegs(context) + REG_SP))
+      if (bump % SIZEOFUINT64 == 0)
+        valid = 1;
 
-  if (bump + size > *(getRegs(context) + REG_SP)) {
-    // out of virtual memory
-    *(getRegs(context) + REG_A0) = 0;
+  if (valid == 0) {
```
ckirsch: Do the else case first.

selfie.c
```diff
@@ -5984,38 +6045,42 @@ void implementMalloc(uint64_t* context) {
...
-
-    // set bump pointer to next free space
-    setBumpPointer(context, bump + size);
   }
 
   setPC(context, getPC(context) + INSTRUCTIONSIZE);
 
-  if (debug_malloc)
-    printf3((uint64_t*) "%s: actually mallocating %d bytes at virtual address %p\n", selfieName, (uint64_t*) size, (uint64_t*) bump);
+  if (debug_brk) {
```
ckirsch: Use `printf2`.

selfie.c
ChristianMoesl: I introduced a function called `emitRoundUp`, because that's really the semantic. 7aa2fa51adf23408a9b3d35480e74fb720d706a8 and 34ec8fe1ccf452fbad23e7e338d5916324193831

selfie.c
ChristianMoesl: 34ec8fe1ccf452fbad23e7e338d5916324193831

selfie.c
ChristianMoesl: 34ec8fe1ccf452fbad23e7e338d5916324193831

selfie.c
ChristianMoesl: 34ec8fe1ccf452fbad23e7e338d5916324193831

selfie.c
ChristianMoesl: 34ec8fe1ccf452fbad23e7e338d5916324193831

selfie.c
ChristianMoesl: 34ec8fe1ccf452fbad23e7e338d5916324193831

selfie.c
ChristianMoesl: 34ec8fe1ccf452fbad23e7e338d5916324193831

selfie.c
ChristianMoesl: 34ec8fe1ccf452fbad23e7e338d5916324193831

selfie.c
ChristianMoesl: 34ec8fe1ccf452fbad23e7e338d5916324193831

selfie.c
ChristianMoesl: 34ec8fe1ccf452fbad23e7e338d5916324193831

ChristianMoesl:
Thanks for the feedback! I polished everything. Should be okay now.

selfie.c
```diff
@@ -4495,6 +4494,18 @@ void compile_cstar() {
 // ------------------------ MACHINE CODE LIBRARY -------------------
 // -----------------------------------------------------------------
 
+void emitRoundUp(uint64_t reg, uint64_t m) {
+  talloc();
+
+  // computes value(reg) -  ((value(reg) + (m - 1)) % m)
```
ckirsch: I am not sure I understand the code. Does this really round up values?

ChristianMoesl: Oh my gosh, you are right. The comment is wrong. I fixed the comment in 3186adf4b088e04129337a7eda054b6e207488b2

First i was irritated that much, that I tried to prove the correctness of this formula. Which of course, is not possible. The good thing is, that I proved now also the correct version, after fixing the comment. 😃

ckirsch:
@ChristianMoesl I am wondering if we can avoid using `$s1` and instead use the first address returned by `brk` to store the bump pointer of `malloc` in memory. Can you try?

ckirsch:
@ChristianMoesl Check indentation of lines 1427-1428.

ChristianMoesl:
@ckirsch Yeah I thought about something like that. The easiest way is, to call `brk` two times in malloc. First to get current brk. Second to set the new brk. Would that be a way to go? I dropped this idea initially because of the constant performance overhead, even if it's cleaner.

The indentation problem, that you mentioned was introduced in the current master branch. But yes, I can fix this. 

Maybe we should also clean the `context` structure, because I think, there is no point for a _bump pointer_ field. It's more like a _minimum program break_, because `brk` semantically really only modifies the program break.

ckirsch:
True. Chicken and egg. But any `malloc`  needs state. How is this usually done? Can you try and find out? Maybe a global variable? Could be implicitly allocated by the compiler as the first global variable before any explicitly declared global variable...

ckirsch:
Yes, please clean up `context` as well by removing the bump pointer field. That field is in fact only necessary for `malloc`, see the discussion `$s1` versus implicit global variable.

ChristianMoesl:
I think you are right. The way to go is, that we use a global variable for the bump pointer. But I think we should give the variable a meaningful name like `__bump` (the underscore variables are reserved for c-library functions/variables). I will implement this and update the PR.

In terms of cleaning up the `context` structure, we still need something like the _minimum program break_ to print the allocated memory when the emulator exits.

ckirsch:
Call it the *original program break*.

ckirsch:
There is one conflict now. I could resolve that but leave it anyway to you to stay in sync. When you are done I pull the PR in.

selfie.c
```diff
@@ -4530,25 +4530,8 @@ void emitStart() {
...
-
-  // set program brk to the aligned bump pointer
-  emitADDI(REG_A7, REG_ZR, SYSCALL_BRK);
-
-  emitECALL();
-
-  // save the actual bump pointer in register s1
-  emitADDI(REG_S1, REG_A0, 0);
-
   // calculate the global pointer value accommodating 6 more instructions
```
ckirsch: The comment needs to be updated to 16.

selfie.c
```diff
@@ -4568,6 +4551,25 @@ void emitStart() {
...
+  // initialize _bump with the original bump pointer
+  emitADDI(REG_A0, REG_ZR, 0);
+  emitADDI(REG_A7, REG_ZR, SYSCALL_BRK);
+
+  emitECALL();
+
+  // align the current bump pointer for double-word access
+  emitRoundUp(REG_A0, SIZEOFUINT64);
+
+  // set program brk to the aligned bump pointer
```
ckirsch: set program break to aligned bump pointer

ckirsch: set program break to aligned bump pointer

selfie.c
```diff
@@ -4568,6 +4551,25 @@ void emitStart() {
     emitADDI(REG_GP, REG_GP, lower);
   }
 
+  // initialize _bump with the original bump pointer
+  emitADDI(REG_A0, REG_ZR, 0);
+  emitADDI(REG_A7, REG_ZR, SYSCALL_BRK);
+
+  emitECALL();
+
+  // align the current bump pointer for double-word access
```
ckirsch: align current bump pointer...

selfie.c
```diff
@@ -4568,6 +4551,25 @@ void emitStart() {
...
+  emitRoundUp(REG_A0, SIZEOFUINT64);
+
+  // set program brk to the aligned bump pointer
+  emitADDI(REG_A7, REG_ZR, SYSCALL_BRK);
+
+  emitECALL();
+
+  // store the bump pointer into the memory
```
ckirsch: store bump pointer in compiler-declared global variable

ChristianMoesl:
Thanks for the feedback again! I did it.

ckirsch:
@ChristianMoesl You still need to resolve the merge conflict please.

ChristianMoesl:
@ckirsch Done!

PR #79, opened 2018-07-01
Simplified docker image creation by importing the riscv tools

ChristianMoesl:
***Problem***
The build process for Spike takes hours. But in order to generate selfie docker images automatically (for every commit on the master branch) by e.g. traffis-ci or the docker hub, the build time needs to be very short. 

***Solution***
I extracted all the code to build the RISC-V tools and just imported this image for this container. This results in very fast builds (< 1min) and enables us to include docker into the travis-ci build pipeline. The downside of this is, that we need to build the RISC-V tools image from time to time manually, if we want an up2date version of Spike.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,85 +1,19 @@
-# builder image
-FROM ubuntu:16.04
+# import image where spike is already installed
+FROM cksystemsteaching/riscv-tools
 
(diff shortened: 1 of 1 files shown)


ckirsch:
Nice! Where is the RISC-V tools image actually stored?

PR #80, opened 2018-07-01
Added Spike as new test target

ChristianMoesl:
Selfie should be automatically tested on spike to enforce compatibility with the RISC-V instruction set. This can be done, by specifying a test target, which get's automatically executed by the CI build pipeline.


Makefile
```diff
@@ -59,6 +59,11 @@ sat: selfie
 	./selfie -sat manuscript/cnfs/rivest.cnf
 	./selfie -c selfie.c -m 1 -sat manuscript/cnfs/rivest.cnf
 
+# Run selfie on Spike
+spike: selfie
```
ckirsch: Please add the spike target to .PHONY

PR #81, opened 2018-07-01
Integrated Spike into build pipeline with the help of docker

ChristianMoesl:
This PR concludes #79 and #80. This patch results in this build structure:
screenshot 2018-07-01 at 23 33 46

**Test on Spike**
A docker image with the current version of selfie gets built. This image will now be run with the arguments `make spike`, which runs selfie on the Spike emulator. The return value of this gets propagated back to the Travis-CI build pipeline.

**Deploy docker image**
Here the docker image will essentially be deployed to the [docker hub](https://hub.docker.com/). The deployed docker image can now be used by every user by typing `docker run -it cksystemsteaching/selfie`, which starts a new container (simplified vm) with the newest version of selfie, spike and the build utils installed. It starts therefore a complete installation with all the things you need to use selfie. Btw: This build stage will only be executed on the master branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ stages:
   - test
   - test on spike
   - name: deploy docker image
-    if: branch = master
+    if: branch = master AND type = push AND fork = false
 jobs:
   include:
     - stage: test on spike
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,28 @@ compiler:
 - clang
 - gcc
 script: travis_wait 60 make all
+stages:
+  - test
+  - test on spike
+  - name: deploy docker image
+    if: branch = master
+jobs:
+  include:
+    - stage: test on spike
+      sudo: required
+      services: docker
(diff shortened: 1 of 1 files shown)


ckirsch:
@ChristianMoesl Can you check this PR and fix it, please?

ChristianMoesl:
@ckirsch I merged it with your branch. To get It working, I set two environment variables in the travis-ci GUI, which are: DOCKER_USERNAME and DOCKER_PASSWORD.

ChristianMoesl:
Okay, the deployment stage will not work here. As stated in their documentation regarding secret environment variables (like DOCKER_PASSWORD) "Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository". So the problem here is, that the deployment to Docker hub is started with an empty password, which gives an error. 

Anyway, it does not make sense either way to deploy builds from a PR. So I will fix the condition for the deployment stage.

PR #89, opened 2018-11-16
Fixing compiler warnings

timung:
Not as hacky as I imagined.
selfie now compiles without warnings using pretty strict warning options (-Wall -Werror).

- Fixed several method signatures to match their built-in versions (exit, malloc, open, ...), renamed some methods (abs, atoi) to avoid the mismatch with built-in methods.
- Commented out some unused methods/variables.
- Added a definition for ulong (unsigned long), which is needed for the correct malloc signature.

int, unsigned long and char* are changed to uint64_t internally.

make all finishes without errors on my machine. (Linux)

The int type might cause some issues down the road, since gcc compiles it to 32 Bit values.
As an example of this, consider this main method: 
```
int main(uint64_t argc, uint64_t* argv) {
  init_selfie((uint64_t) argc, (uint64_t*) argv);

  init_library();
  exit(two_to_the_power_of(32));
  return selfie();
} 
```
Running echo $? afterwards reveals an exit code of 0, since the exit code is of type int, while using `exit(two_to_the_power_of(32) - 1);` correctly produces a non-zero exit status.


ckirsch:
Thanks a lot, @timung!

The PR contains a lot of unrelated file changes but it did inspire me to change bootstrapping in such a way that there are no compiler warnings anymore, at least on Travis.

PR #90, opened 2018-12-12
Align p_offset in ELF to page size, and use standard openat syscall.

joanbm:
If I understand correctly, in ELF files, the `p_offset` field must be a multiple of `p_align`. See in man elf(5) ( read at http://man7.org/linux/man-pages/man5/elf.5.html ):

```
       p_align   This member holds the value to which the segments are
                 aligned in memory and in the file.  Loadable process seg‐
                 ments must have congruent values for p_vaddr and p_offset,
                 modulo the page size.  Values of zero and one mean no
                 alignment is required.  Otherwise, p_align should be a pos‐
                 itive, integral power of two, and p_vaddr should equal
                 p_offset, modulo p_align.
```

Emphasis on: "Loadable process segments must have congruent values for p_vaddr and p_offset, modulo the page size".

This pull request fixes this, by introducing padding between the ELF header and the code segment. 

I found this, by trying to run selfie-generated executable under QEMU. Without this fix, if one attempts to load a selfie-generated executable with `qemu-riscv64`, it throws an obscure "Invalid argument" error. After some debugging, I found that QEMU is trying to `mmap` the segment and it fails if the segment is not aligned to the page size (4KiB for RISC-V). (I guess that it works in spike with the pk kernel because it is more permissive.)

With this fix, I can run selfie with QEMU:

```
[zealcharm@SOL ~/selfie]$ make
cc -Wall -Wextra -O3 -m64 -D'uint64_t=unsigned long long' selfie.c -o selfie
[zealcharm@SOL ~/selfie]$ ./selfie -c selfie.c -o selfie.elf
./selfie: selfie compiling selfie.c with starc
./selfie: 292655 characters read in 10145 lines and 1356 comments
./selfie: with 172712(59.01%) characters in 44285 actual symbols
./selfie: 347 global variables, 442 procedures, 420 string literals
./selfie: 2543 calls, 1156 assignments, 89 while, 882 if, 397 return
./selfie: symbol table search time was 2 iterations on average and 49390 in total
./selfie: 176624 bytes generated with 40980 instructions and 12704 bytes of data
./selfie: init:    lui: 2849(6.95%), addi: 13708(33.45%)
./selfie: memory:  ld: 7189(17.54%), sd: 5951(14.52%)
./selfie: compute: add: 3987(9.72%), sub: 707(1.72%), mul: 816(1.99%), divu: 82(0.20%), remu: 35(0.08%)
./selfie: control: sltu: 630(1.53%), beq: 975(2.37%), jal: 3601(8.78%), jalr: 442(1.07%), ecall: 8(0.01%)
./selfie: 180720 bytes with 40980 instructions and 12704 bytes of data written into selfie.elf
[zealcharm@SOL ~/selfie]$ chmod +x selfie.elf
[zealcharm@SOL ~/selfie]$ qemu-riscv64 selfie.elf
selfie.elf: usage: selfie { -c { source } | -o binary | [ -s | -S ] assembly | -l binary | -sat dimacs } [ ( -m | -d | -r | -n | -y | -min | -mob ) 0-64 ... ]
```

PS: It still doesn't work completely under QEMU, because the `open` system call is broken. From what I've gathered, the open with syscall number 1024 is deprecated and something else should be used. But that's better handled in another PR...


joanbm:
I did a bit of research on the open syscall number situation, and it turns out to be easy to fix, so I went ahead and added it to the PR.

As far as I can tell, Linux doesn't have an open syscall, but rather an openat syscall. The library's open(...) is equivalent to the library's openat(AT_FDCWD, ...) as pointed by man open(2), which ends up directly translating to the syscall params.

Some resources to confirm this:

* Linux's generic syscalls: https://github.com/torvalds/linux/blob/master/include/uapi/asm-generic/unistd.h ; you can only see an openat syscall

* Linux's RISCV-specific syscalls: https://github.com/torvalds/linux/blob/master/arch/riscv/include/uapi/asm/unistd.h ; there's only a flush icache specific syscall, no open syscall

* Glibc's open: https://code.woboq.org/userspace/glibc/sysdeps/unix/sysv/linux/open.c.html ; it ends up doing an openat syscall

As far as I can tell, the open syscall with number 1024 is something non-standard or deprecated, here's the declaration in Spike PK: https://github.com/riscv/riscv-pk/blob/master/pk/syscall.h ; it's under some `OLD_SYSCALL_THRESHOLD` constant.

What I did in selfie's code, is changing the implementation of open to use the openat syscall instead (with the extra AT_FDCWD parameter). open's signature doesn't change, so it's a pretty self-contained change.

With this, I've found no problems running selfie's RISC-V executable under qemu-riscv64. It's pretty convenient since QEMU is both more available in Linux distros and far more performant than spike. Maybe it's also possible that this allows running selfie's RISC-V executable on real hardware...?

ChristianMoesl:
Hi, sry for the late response. Really good work! Especially seeing selfie running on QEMU is really great.

But I think we can simplify the code a little bit, by increasing the ELF_HEADER_LEN to PAGESIZE, such that we can write/read the header, code length and the padding bytes with one write/read command. May you do that?

I will merge it in, as soon as you change the code!

joanbm:
> Hi, sry for the late response. Really good work! Especially seeing selfie running on QEMU is really great.
> 
> But I think we can simplify the code a little bit, by increasing the ELF_HEADER_LEN to PAGESIZE, such that we can write/read the header, code length and the padding bytes with one write/read command. May you do that?
> 
> I will merge it in, as soon as you change the code!

Thank you for taking a look!

Good idea about unifying the read/write handling of the various header fields&padding into the same call. I've updated the code a bit to redefine ELF_HEADER_SIZE to 4096, so that the ELF header handling code also manages the code_length field and the padding up to the page size, and I think it looks good, since it simplifies the logic a bit. I've updated the PR as well.

Makefile
```diff
@@ -80,4 +80,4 @@ clean:
 	rm -rf *.m
 	rm -rf *.s
 	rm -rf selfie
-	rm -rf selfie.exe
```
ckirsch: Please avoid introducing trailing whitespace.

selfie.c
```diff
@@ -889,7 +889,7 @@ uint64_t MAX_BINARY_LENGTH = 262144; // 256KB = MAX_CODE_LENGTH + MAX_DATA_LENGT
 uint64_t MAX_CODE_LENGTH = 245760; // 240KB
 uint64_t MAX_DATA_LENGTH = 16384; // 16KB
 
-uint64_t ELF_HEADER_LEN = 120; // = 64 + 56 bytes (file + program header)
```
ckirsch: This comment should not be lost. I suggest to keep it but rephrase it to e.g.:

// page-aligned ELF header with sufficient space to store 64 bytes for file header and 56 bytes for program header

selfie.c
```diff
@@ -955,11 +955,13 @@ uint64_t debug_write = 0;
...
-uint64_t SYSCALL_WRITE = 64;
-uint64_t SYSCALL_OPEN  = 1024;
-uint64_t SYSCALL_BRK   = 214;
+uint64_t SYSCALL_EXIT   = 93;
+uint64_t SYSCALL_READ   = 63;
+uint64_t SYSCALL_WRITE  = 64;
+uint64_t SYSCALL_OPENAT = 56;
+uint64_t SYSCALL_BRK    = 214;
+
+uint64_t DIRFD_AT_FDCWD = -100;
```
ckirsch: We could use a comment here explaining that constant.

selfie.c
```diff
@@ -5512,12 +5514,12 @@ void emit_data_segment() {
...
 }
 
-uint64_t* create_elf_header(uint64_t binary_length) {
+uint64_t* create_elf_header(uint64_t binary_length, uint64_t code_length) {
   uint64_t* header;
 
   // store all numbers necessary to create a minimal and valid
   // ELF64 header including the program header
-  header = smalloc(ELF_HEADER_LEN);
+  header = touch(zalloc(ELF_HEADER_LEN), ELF_HEADER_LEN);
```
ckirsch: That doesn't make sense. zalloc zeroes the allocated space. Touching is therefore subsumed by that.

selfie.c
```diff
@@ -5541,24 +5543,30 @@ uint64_t* create_elf_header(uint64_t binary_length) {
...
-  *(header + 9)  = ELF_HEADER_LEN + SIZEOFUINT64; // segment offset in file
+  *(header + 9)  = ELF_HEADER_LEN;                // segment offset in file (must be page-aligned)
   *(header + 10) = ELF_ENTRY_POINT;               // virtual address in memory
   *(header + 11) = 0;                             // physical address (reserved)
   *(header + 12) = binary_length;                 // size of segment in file
   *(header + 13) = binary_length;                 // size of segment in memory
   *(header + 14) = PAGESIZE;                      // alignment of segment
 
+  // This field is not really part of the ELF header, but rather internally
+  // used by Selfie to load its own generated ELF files
```
ckirsch: Selfie -> selfie

selfie.c
```diff
@@ -5729,28 +5729,21 @@ void selfie_load() {
...
+          // check if we are really at EOF
+          if (read(fd, binary_buffer, SIZEOFUINT64) == 0) {
+            printf5((uint64_t*) "%s: %d bytes with %d instructions and %d bytes of data loaded from %s\n",
+              selfie_name,
+              (uint64_t*) ELF_HEADER_LEN,
+              (uint64_t*) (code_length / INSTRUCTIONSIZE),
+              (uint64_t*) (binary_length - code_length),
+              binary_name);
+
+            return;
```
ckirsch: Nice!

selfie.c
```diff
@@ -10121,4 +10116,4 @@ int main(int argc, char** argv) {
   init_library();
 
   return selfie();
-}
\ No newline at end of file
+}
```
ckirsch: See Makefile.

selfie.c
```diff
@@ -6178,18 +6173,18 @@ void implement_open(uint64_t* context) {
...
-    print_register_value(REG_A0);
+    print_register_value(REG_A1);
   }
 
-  vfilename = *(get_regs(context) + REG_A0);
-  flags     = *(get_regs(context) + REG_A1);
-  mode      = *(get_regs(context) + REG_A2);
+  vfilename = *(get_regs(context) + REG_A1);
+  flags     = *(get_regs(context) + REG_A2);
+  mode      = *(get_regs(context) + REG_A3);
```
ckirsch: We need a comment here why A0 is ignored.

ckirsch:
@joanbm @ChristianMoesl What do you think about including Qemu for RISC-V in our docker image? Is that difficult to do?

selfie.c
joanbm: @ckirsch Actually, this (or something similar) is necessary, since at boot levels >0, zalloc falls back to malloc, which doesn't guarantee that the memory is mapped. If you remove this call to touch and run 'make all', the test fail for this reason. I have kept the call to touch and added a clarification comment, though I'm not sure if there's a better solution.

joanbm:
> @joanbm @ChristianMoesl What do you think about including Qemu for RISC-V in our docker image? Is that difficult to do?

Since support for RISC-V in QEMU was only upstreamed relatively recently, it isn't (and most likely won't be) available on Ubuntu 16.04 LTS, so it's not as easy as adding some extra packages to apt-get. It also [isn't available on Ubuntu 18.04 LTS](https://packages.ubuntu.com/bionic/amd64/qemu-user/filelist), but [it is on Ubuntu 18.10 (non-LTS)](https://packages.ubuntu.com/cosmic/amd64/qemu-user/filelist).

Anyway, it shouldn't be too hard to get it working - either by switching the image to another distribution that has it (which probably won't be an LTS), by building it from source, or by grabbing a prebuilt binary from a qemu-user-static package. Not sure what's the best option or if it's worth it, though.

selfie.c
ckirsch: Interesting! Let me suggest to clean this up a bit more. There are two code locations in which ELF headers are allocated. One in which zalloc is used, and one in which smalloc is used. Let's have instead just one short function called allocate_ELF_header which uses zalloc and touch.

joanbm: Done - If I'm not missing anything, all requested code changes are done

joanbm:
@ckirsch I have made an initial attempt of building qemu from the riscv-tools repository within the Docker image, seems to work well on some basic tests. I've pushed the changes to GitHub (in a separate branch), and also pushed a prebuilt riscv-tools with qemu to Docker Hub, in case you want to take a look or play with it.

https://github.com/joanbm/selfie32/commit/04aaceca0803f883ad1bbe4040a116052abd4829

https://hub.docker.com/r/joanbm/selfie-riscv64-tools-with-qemu

ckirsch:
@joanbm This is great! Thanks a lot. The only problem is that I don't have time right now to look at this. @ChristianMoesl Can you please take a look and see if we can integrate that? Ideally, we should have support of spike/pk and QEMU/Linux in a single docker image, and announce that in the README.md

PR #91, opened 2019-01-22
Fix uninitialized memory and out of bounds reads when working with strings

joanbm:
When doing some tests with the Dockerfiles related to my last pull request, I found some problems ultimately caused with memory safety issues in selfie, when working with strings. Those can be seen and diagnosed with valgrind.

1. In some places (e.g. when allocating an identifier string, see `MAX_IDENTIFIER_LENGTH`), memory is allocated with `smalloc`, which does not initialize it. When calling `store_character`, this does an uninitialized read. (On modern systems, this does not cause a crash as it could, but rather read some leftover value).

2. Similarly, in most of those same places, the allocated memory size is not a power of 8 (`SIZEOFUINT64`). This may cause an out of bounds read when calling `store_character`. (On modern systems, this does not cause a crash as it could, because `malloc` tends to add some alignment/padding to fix this).

3. Finally, I found a place where the previous issues are not merely theoretical/obscure, but rather does actually cause selfie not to work on realistic tests. When allocating an identifier, memory is allocated with `smalloc`, which does not necessarily zero-initialize it. Later, `hash` (used to build the symbol tables) can read a whole double word from this identifier, which may have some uninitialized garbage at the end. Different invocations of `hash` on different allocations/instances of the same string can read different garbage, so the hash can be different, which can cause compilation to fail.

I can 100% reproduce this last problem with an Alpine Linux Docker container (probably because musl's malloc is less generous than glibc?). Not sure if YMMV:

```
[zealcharm@SOL ~/selfietest]$ cat Dockerfile-test
FROM alpine:3.8
RUN apk update && apk add gcc musl-dev make
COPY . /opt/selfie/
WORKDIR /opt/selfie
RUN make selfie
CMD /bin/ash
[zealcharm@SOL ~/selfietest]$ sudo docker build . -f Dockerfile-test -t selfietest && sudo docker run --rm -it selfietest
[...]
/opt/selfie # ./selfie -c selfie.c
./selfie: selfie compiling selfie.c with starc
./selfie: syntax error in selfie.c in line 6721: rs2 undeclared
```

This PR adds a `string_alloc` function that makes sure string allocations are a multiple of `SIZEOFUINT64` and are initially zeroed out to fix the above problems.

PS: We could technically avoid manually adding the string terminator since the string is already zeroed, not sure if it helps readability, though.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -127,6 +127,7 @@ uint64_t sign_shrink(uint64_t n, uint64_t b);
 uint64_t  load_character(uint64_t* s, uint64_t i);
 uint64_t* store_character(uint64_t* s, uint64_t i, uint64_t c);
 
+uint64_t* string_alloc(uint64_t l);
 uint64_t  string_length(uint64_t* s);
 uint64_t* string_copy(uint64_t* s);
 void      string_reverse(uint64_t* s);
@@ -274,7 +275,7 @@ void init_library() {
   *character_buffer = 0;
 
   // accommodate at least CPUBITWIDTH numbers for itoa, no mapping needed
-  integer_buffer = smalloc(CPUBITWIDTH + 1);
+  integer_buffer = string_alloc(CPUBITWIDTH);
 
   // does not need to be mapped
   filename_buffer = smalloc(MAX_FILENAME_LENGTH);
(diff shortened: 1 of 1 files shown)


joanbm:
Also a good idea to add a CI test that checks that Valgrind reports no safety errors on some test cases.

ckirsch:
This is great! Thanks a lot, @joanbm 

I was wondering if you could try and add a (platform-independent) valgrind target to the Makefile and do a PR on that.

Also, the discussion on QEMU needs to be extended. We also need boolector installed in that docker image along with spike/pk. This is because we are now generating SMT-LIB files that should be checked by boolector.

ckirsch:
@joanbm That's much appreciated! Thanks!

Please have a look at https://github.com/cksystemsteaching/selfie/commit/94b7a0702eac29984d26fbe9e89e3fbe80d3fc6a

I changed the `string_alloc` code to use smalloc rather than zalloc but kept the alignment code, which on boot level zero is indeed necessary. My overall reasoning is that, if selfie is correct, such memory should always be initialized properly without zeroing it. If not, I am hoping that valgrind will find the issue.

ckirsch:
Oops, not true. Never mind. Strings are emitted into the data segment, including any uninitialized bytes past the null terminator. So changed `string_alloc` back to zeroing.

PR #92, opened 2019-02-04
Updated the monster function

Assklou:
remove the uint64_t b feature.


selfie.c
```diff
@@ -9592,4 +9589,4 @@ int main(int argc, char** argv) {
   init_library();
 
   return selfie();
-}
\ No newline at end of file
+}
```
ckirsch: @Assklou Please remove the trailing linefeed. Thanks!

PR #95, opened 2019-02-14
Native Threads

mwidmoser:
This PR is the outcome of the "native" thread implementation developed during the Corey project in SWS. 

## Overview
* Introduced some new system calls (`fork`, `thread`, `wait`, `mmap`, `pid`, `lock`, `unlock`).
* Simple locking mechanism by using `__sync_bool_compare_and_swap()`, an atomic built-in provided by gcc.
* Native fork and wait support for processes.
* Imitating threads by allocating and assigning shared memory with mmap.
* Test file.

## Problems
* We added a new Makefile target, could be that Travis will fail since I am not sure if running more than one process is allowed on their servers.

* Unfortunately, compiling with gcc and all warnings turned on gives a warning. 
This is due to a type conflict of the built-in procedure `fork()`. In our implementation, `fork()` returns the type `uint64_t`, however, `fork()` by default returns an integer (signed) of type `pid_t`. A simple solution would be renaming `fork`.
    ```
    cc -Wall -Wextra -O3 -m64 -D'uint64_t=unsigned long long' selfie.c -o selfie
    selfie.c:117:10: warning: conflicting types for built-in function ‘fork’ [enabled by default]
    uint64_t fork();
            ^
    ```

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -110,6 +110,10 @@ uint64_t open(char* filename, uint64_t flags, uint64_t mode);
 // selfie bootstraps void* and unsigned long to uint64_t* and uint64_t, respectively!
 void* malloc(unsigned long);
 
+uint64_t fork();
+uint64_t wait();
+uint64_t getpid();
+
 // -----------------------------------------------------------------
 // ----------------------- LIBRARY PROCEDURES ----------------------
 // -----------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


mwidmoser:
The Travis CI build for target `Test` actually passes, however, for `Spike` it fails. The new system calls are not supported natively, except for mmap.

selfie.c
```diff
@@ -4679,11 +4833,11 @@ void compile_cstar() {
...
 
 void emit_round_up(uint64_t reg, uint64_t m) {
-  talloc();
+  load_integer(m - 1);
 
   // computes value(reg) + m - 1 - (value(reg) + m - 1) % m
-  emit_addi(reg, reg, m - 1);
-  emit_addi(current_temporary(), REG_ZR, m);
+  emit_add(reg, reg, current_temporary());
+  emit_addi(current_temporary(),current_temporary(), 1);
```
ckirsch: One space after the first `,`, please.

ckirsch:
I am not sure if you actually need `sfork`. Can this be avoided?

saraseidl:
We need sfork to distinguish between thread and fork internally. We thought this to be the cleanest way.

Both thread and fork are implemented using fork on bootlevel, but additional steps are required for 'real' fork. 

sfork is used only internally for the implementation of thread and does not copy shared space on bootlevel 0.
sfork so to say forks the emulators 'below'. 

fork uses fork in its implementation and does copy shared space on bootlevel 0.

selfie.c
mwidmoser: Done.

mwidmoser:
> @mwidmoser @saraseidl Could you please do the PR against the `concurrency` branch? I just created that.

I changed the base branch to `concurrency`. I also commented out the invocation of the spike emulator in the Makefile-target for now. Travis should pass then.

PR #97, opened 2019-02-19
(RFC/Draft) qemu/boolector automated tests and Dockerfile build time improvements

joanbm:
This is a proposal for various changes to the Dockerfiles used by selfie to deploy the Docker image that is used for automated testing and easy deployment. A lot of those changes are opinionated/debatable and it's possible I am missing something, so I am opening this PR as a draft / request for comments first.

Summary of the changes:
* **`Dockerfile`**: Do not compile the entire RISCV compile toolchain just to compile `pk`, but rather rely on the RISCV GCC/libc on Ubuntu's 18.10 repositories. This saves a lot of time when building the Docker image. The downside that I can see are that the Ubuntu 18.10 base image, which is not an LTS release.
* **`Dockerfile`**: Download only the required repositories, not all submodules from riscv-tools. Again, this greatly improves build time.
* **`Dockerfile`**: Parallelize build with make. Improves build times.
* **`Dockerfile`**: Add QEMU executable to the final image. For simplicity, the `qemu-riscv64-static` binary from Ubuntu's 18.10 repositories is taken as well, but is copied into the final image (which is still an Ubuntu 16.04 LTS image, i.e. has not changed).
* **`Dockerfile`**: Add boolector compilation stage and add the resulting binaries to the final image. Note that even though boolector is in Ubuntu's 18.10 repositories, the available version is too old to be used in selfie's case. Therefore, it is built from source.
* **`Dockerfile`**: Split the build into multiple stages (Proxy Kernel, Spike, QEMU, Boolector, final selfie image). This facilitates better isolation between the stages (which also translates into improved use of caches, and hence faster build times, for repeated rebuilds with slight changes).
* **`Dockerfile` and `Makefile`**: Remove `Dockerfile-riscv-tools` Docker image, and only use a single `Dockerfile` with all the stages. Because of the previous changes, a full rebuild takes ~5 minutes on a typical computer, or ~15 minutes on Docker Hub's automated builds, so I don't believe the added complexity is worth it.
* **`Makefile`**: Add test of selfie using qemu. This test is analogous to the existing one using Spike.
* **`Makefile`**: Add test of selfie using boolector. The test I've implemented is simple but not very useful, and can certainly be improved, but it is enough to check that selfie's SAT is working and returning credible results.
* **`travis.yml`**: Added qemu and boolector tests to the Travis CI automatic test.


ckirsch:
@joanbm Thanks a lot! This looks great. @ChristianMoesl What do you think? You developed the original docker image. I'd be glad to accept the PR here.

ckirsch:
Thanks a lot @joanbm

Your work is much appreciated!

PR #101, opened 2019-03-22
Risc v to x86

alexkollert:


selfie.c
```diff
@@ -9572,6 +10752,14 @@ uint64_t selfie() {
         return selfie_run(MONSTER);
       else if (string_compare(option, "-y"))
         return selfie_run(HYPSTER);
+      else if (string_compare(option, "-t")) {
```
ckirsch: This should go into a procedure called `selfie_translate`.

selfie.c
```diff
@@ -8103,6 +9249,38 @@ void print_profile() {
   }
 }
 
+void translate_to_assembler() {
```
ckirsch: This should be renamed to `translate_to_RISC_V_assembler`.

selfie.c
```diff
@@ -7632,366 +8817,325 @@ void fetch() {
...
+    constrain_jalr();
+    do_jalr();
+  } else if (is == LUI) {
+    constrain_lui();
+    do_lui();
+  } else if (is == ECALL)
+    do_ecall();
+}
+
+void execute_translate() {
```
ckirsch: I am assuming this is not execution, right? You only go through the code once and translate each instruction, right? If yes, I'd call this procedure `translate_to_x86_binary`.

ckirsch: I am using the prefix `execute_` only for procedures that actually execute code in some way.

selfie.c
ckirsch: Let's use `-x86` as option.

ckirsch: Please prepare an (optional, not among all) Makefile target `x86` that tests self-compilation using the x86 binary you generate. Then, add that target to the test targets of the docker image in travis.

ckirsch:
@alexkollert Let me also recreate the dev branch here to be up-to-date with master. Please pull those changes in before working on my feedback. Thanks!

PR #119, opened 2019-04-10
Fix incompatibility between RISC-U and RISC-V assembly files

ChristianMoesl:
The grader already checks, if students implement the assembly instructions, such that they are compatible with RISC-V GCC. So I thought, that the assembly file generated with Selfie's `-s` option, should also be compatible with the RISC-V toolchain. This commit only changes the output for the `-s` option. The assembly file generated with `-S` stays the same, with the exception of the new register names. Note, that I had to change the name for the `fp` register to `s0`.

current:
```
0x13C: beq $a0,$t1,2[0x100]
0x140: sd $fp,0($sp)
0x144: .quad 0x626F6D2D
```

new:
```
beq a0,t1,2
sd s0,0(sp)
.quad 0x626F6D2D
```


ckirsch:
This is very nice! I didn't expect that RISC-U assembly was so close to official RISC-V assembly.

Could you please add an optional target to the Makefile that checks if selfie generated assembly is in fact properly formatted and then perform that check in the docker image where the tools should be available?

ChristianMoesl:
I am on it! 👍

PR #123, opened 2019-04-18
Added test files for btor2

ChrisEdel:


ckirsch:
Thanks! Please change the path to the symbolic examples in the clean target of the Makefile as well.

PR #127, opened 2019-04-30
updating manuscript/code/hello-world example(s) with recent c* code

OxiAction:
- no more warnings
- char* instead of uint64_t*

### DIFF
--- a/manuscript/code/hello-world-minified.c
+++ b/manuscript/code/hello-world-minified.c
@@ -1,1 +1,1 @@
-uint64_t*foo;uint64_t*main(){foo="Hello World!    ";while(*foo!=0){write(1,foo,8);foo=foo+1;}}
+char*foo;uint64_t main(){foo="Hello World!    ";while(*foo!=0){write(1,foo,8);foo=foo+1;}}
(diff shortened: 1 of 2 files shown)


manuscript/code/hello-world-minified.c
```diff
@@ -1 +1 @@
-uint64_t*foo;uint64_t*main(){foo="Hello World!    ";while(*foo!=0){write(1,foo,8);foo=foo+1;}}
\ No newline at end of file
+char*foo;uint64_t main(){foo="Hello World!    ";while(*foo!=0){write(1,foo,8);foo=foo+1;}}
```
ckirsch: This breaks the original idea of zero whitespace in the minified version...

OxiAction: changed it back to `uint64_t*main`

ckirsch:
@OxiAction Your input is appreciated but your PR still won't make a difference here, and the warning from selfie is still there. The sole purpose of supporting the `char` keyword is to make `gcc/clang` stay silent during bootstrapping selfie. However, the hello world programs do not require bootstrapping.

PR #134, opened 2019-05-21
Fix page table caching of vctxt for stack pages

### LINKED ISSUE #133: Hypster crashes with many local variables
The following code crashes when executed with Hypster with a error message "selfie.c: palloc out of physical memory".

`./selfie -c selfie.c -m 128 -c code.c -y 64`

code.c:
```
int main(int argc, char** argv) {
  uint64_t a0;
  uint64_t a1;
  // .
  // .
  // .
  uint64_t a1099;
  uint64_t a1100;

  a1100 = 1;

  return a1100;
}
```

ChristianMoesl:
This PR fixes the issue #133 with caching of stack pages in the page table. The previous caching algorithm in `restore_context()` relies on the fact, that the stack grows continuously downwards in the address space and ignores the fact, that there could be stack pages, which do not get mapped. A simple example for this is something like this:

```
void f() {
  uint64_t a0;
  // ...
  uint64_t a9999;
  uint64_t a10000;

  a10000 = 1;
```

Clearly, as `a10000 = 1` get's executed, a page `n` get's mapped, whereas the page `n + 1` is not mapped. In this situation, the current caching algorithm will fail, because it relies on the fact, that all pages from the highest virtual-memory address until stack-top are mapped.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -1433,33 +1433,34 @@ uint64_t* delete_context(uint64_t* context, uint64_t* from);
 // |  2 | program counter | program counter
 // |  3 | registers       | pointer to general purpose registers
 // |  4 | page table      | pointer to page table
-// |  5 | lo page         | lowest low unmapped page
-// |  6 | me page         | highest low unmapped page
-// |  7 | hi page         | highest high unmapped page
-// |  8 | original break  | original end of data segment
-// |  9 | program break   | end of data segment
-// | 10 | exception       | exception ID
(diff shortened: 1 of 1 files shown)


ChristianMoesl:
I refactored my code as you suggested. But I did choose an other name for this procedure, because `map_region` does not fit the semantics, I think. This procedure really does only copy allready mapped frames into the page table of a cache-context and therefore restores the pt region.

PR #135, opened 2019-06-06
Implemented follow and merge decision

ChrisEdel:


selfie.c
```diff
@@ -8113,9 +8612,54 @@ void execute_symbolically() {
...
     do_jal();
-  else if (is == JALR) {
+
+    pc_after_jal = pc;
+
+    // we need to check if we jump into a recursion
+    fetch();
+    decode();
+
+    if (is == ADDI) {
```
ckirsch: You can simplify this by just checking if the JAL uses the RA or the ZERO register. In the former case it is a procedure call.

ckirsch: Just search for `emit_JAL` to see that.

selfie.c
```diff
@@ -7765,6 +7888,379 @@ char* smt_binary(char* opt, char* op1, char* op2) {
...
+  }
+
+  // we need to check if we are inside of a recursion before we reach the merge location
+  // if we are, we merge only when the recursion is finished
+  while (pc != merge_location) {
+    fetch();
+    decode();
+
+    if (is == JAL)
+      if (is_start_of_procedure_prologue(pc + imm)) {
```
ckirsch: Same here. Check if JAL uses RA register.

selfie.c
```diff
@@ -7677,6 +7752,54 @@ void store_symbolic_memory(uint64_t vaddr, uint64_t val, char* sym, char* var, u
...
+}
+
+void delete_word_from_symbolic_memory(uint64_t vaddr) {
+  uint64_t* sword;
+
+  sword = symbolic_memory;
+
+  while (sword) {
+    if (get_word_address(sword) == vaddr)
+      // note: we do not really delete the word, we just set the virtual address to -1
```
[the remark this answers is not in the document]
ChrisEdel: Here it should be DELETED and somewhere else MERGED, I have introduced both constants.

selfie.c
ChrisEdel: That is much better! I did not think of that, I changed it.

selfie.c
```diff
@@ -8819,18 +9381,26 @@ uint64_t handle_timer(uint64_t* context) {
   set_exception(context, EXCEPTION_NOEXCEPTION);
 
   if (symbolic) {
-    print("(push 1)\n");
+    print(";(push 1)\n");
```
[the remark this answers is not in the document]
ChrisEdel: Yes, that is not necessary. I removed that.

selfie.c
ChrisEdel: Actually the name of the procedure is a bit misleading, it is not only enough to check if we call a procedure, we also have to check that we are currently in this procedure which we call. I have improved that.

selfie.c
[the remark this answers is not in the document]
ckirsch: But where do you push? Even if you always merge you still get more than one formula to solve, right?

Makefile
```diff
@@ -88,8 +88,7 @@ qemu: selfie.m selfie.s
 # Test boolector SMT solver
 boolector: smt
 	boolector manuscript/code/symbolic/simple-assignment.smt -e 0 > selfie_boolector.sat
-	[ $$(grep ^sat$$ selfie_boolector.sat | wc -l) -eq 2 ]
-	[ $$(grep ^unsat$$ selfie_boolector.sat | wc -l) -eq 1 ]
```
ckirsch: The unsat case is still possible, right? Why not check for that?

selfie.c
```diff
@@ -7657,6 +7736,9 @@ void store_symbolic_memory(uint64_t vaddr, uint64_t val, char* sym, char* var, u
 
   sword = allocate_symbolic_memory_word();
 
+  // we 'delete' the word, if it already exists in the symbolic memory
```
ckirsch: We mark it as deleted...

selfie.c
```diff
@@ -7677,6 +7759,58 @@ void store_symbolic_memory(uint64_t vaddr, uint64_t val, char* sym, char* var, u
...
+}
+
+void delete_word_from_symbolic_memory(uint64_t vaddr) {
+  uint64_t* sword;
+
+  sword = symbolic_memory;
+
+  while (sword) {
+    if (get_word_address(sword) == vaddr)
+      /* we indicate that this word has been deleted in order to avoid finding
```
ckirsch: I remember that I solved this problem by simply adding new values of a word to the front of a list of values. Later, when looking for the value of a word I searched from the beginning of the list and then stopped when I found the first value. Would that work here as well?

selfie.c
```diff
@@ -7765,6 +7899,421 @@ char* smt_binary(char* opt, char* op1, char* op2) {
...
+        merge_location = recursive_merge_location;
+        in_recursion = 1;
+      }
+
+    pc = pc + INSTRUCTIONSIZE;
+  }
+
+  // restore the original program state
+  pc = original_pc;
+  imm = original_imm;
```
ckirsch: You don't need to remember `imm`. Decode will restore that, no?

selfie.c
```diff
@@ -7765,6 +7899,421 @@ char* smt_binary(char* opt, char* op1, char* op2) {
...
+    return;
+  }
+
+  print("; merging two contexts at ");
+  print_code_context_for_instruction(location);
+  println();
+
+  if (prologues_and_corresponding_merge_locations != (uint64_t*) 0)
+    if (get_pc(active_context) == *(prologues_and_corresponding_merge_locations + 2))
+      // we have finished the recursion (i.e. the program has reached a program location which is not part of any recursion)
```
ckirsch: I don't see why you can't merge in some recursion. Can you explain?

selfie.c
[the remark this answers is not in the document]
ChrisEdel: I push only when a path does properly exit, if a path is reaches its maximal execution depth (i.e. is timed out), I do not push since this path did not properly finish execution.
Normally all paths are merged, this means in particular that the last if statement in the examples is also merged and there is overall only one formula to solve.

Makefile
ChrisEdel: The last if statement is also merged in the example, which means that only one path remains at the end, and thus only one formula remains to be solved, which is satisfiable in this case, so the unsat case no longer exists.

selfie.c
ChrisEdel: At the moment I have implemented it in such a way that the merging always happens outside of the entire recursion (if there is any). It is of course possible to merge within a recursion, but you have to be very careful, because you can only merge on the same recursion level (because the paths have to be at exactly the same location).  And the problem is that recursions can be nested arbitrarily, so it is difficult to keep track of which paths are on the same recursion level and could be merged. I have been thinking about this for a while and I think it would take quite a lot of code so that every context knows how deep it is in the recursion. Of course it would be more efficient to merge as soon as possible considering the translation time, but for the formula at the end it should not make much of a difference if you merge a bit later (i.e. outside the recursion) since the merging happens anyway. So the question is whether an earlier merge is worth it to make the engine (much?) more complex.

ChrisEdel: Actually I found an issue today regarding the recursion during some tests. I already know how to fix this, I will take care of it and think about it again if we can/should merge earlier in the recursion.

selfie.c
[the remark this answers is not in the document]
ckirsch: But what if there is an error on a path where you reach the maximal execution depth? Say there is a division-by-zero error on such a path and no other path. In that case, you'd miss that, unnecessarily, no?

Makefile
ckirsch: Got it!

selfie.c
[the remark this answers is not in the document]
ChrisEdel: The division-by-zero error would cause the path to stop executing and would result in an exit (the method handle_division_by_zero returns EXIT), so this path would exit 'properly' and would be pushed. So nothing should be missed, or am I looking at this the wrong way?

selfie.c
ckirsch: Yes, please do. This is an interesting problem! Knowing if you are at a mergeable location appears to involve keeping track of the RA register values on the stack. I believe two locations can only be merged if they share the same prefix of RA register values.

selfie.c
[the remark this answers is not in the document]
ChrisEdel: My previous answer is not correct. I can see what you mean now. I will have to take a look at it, I will let you know.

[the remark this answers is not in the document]
ckirsch: Division by zero is only handled during concrete execution on mipster level. Your goal is to encode binaries in such a way that, using an SMT solver, you can compute input that will make the binaries do a division by zero during concrete execution. The same is true for other errors such as unsafe memory access.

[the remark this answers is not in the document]
ckirsch: So, not pushing timed-out paths misses such errors if they occur on parts of the timed-out paths that are not merged into other paths for which formulae are generated.

selfie.c
ckirsch: Is this true?

selfie.c
ckirsch: Also here. Is this true?

selfie.c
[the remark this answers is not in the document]
ChrisEdel: That is true, I will work on that.

selfie.c
ChrisEdel: That is true, I was just going to fix that in the next update.

selfie.c
ChrisEdel: I do not think so, but I am not quite sure yet. I am currently working on it and will tell you as soon as I am sure and have a good reasoning.

ChrisEdel: In `merge_symbolic_store` I have to go through the symbolic memory of both contexts to merge words with the same virtual address. That means I would have to remember (at least in the outer loop) which words (or which virtual addresses) I have already looked at, so this would lead to something similar and I thought that it would be a bit clearer to mark such words as DELETED right from the beginning.

ckirsch: I believe in that case it is easier and temporally and spatially more efficient to maintain the invariant that there is at most one entry per vaddr in any symbolic store. This means whenever you create an entry for a given vaddr you need to search the symbolic store. If you find an entry for vaddr just update that entry and you are done, thereby avoiding any duplicate entries. If you do not find an entry create a new entry in the store. The store is a singly-linked list with new entries added at the head of the list and existing entries searched by also starting at the head of the list. Despite quadratic complexity, this should facilitate fast symbolic storing with code that exhibits temporal locality (a store to a given vaddr is likely followed by another store to the same address) which most code does. Overall, the merge should then become very simple, right?

ckirsch: Actually, I am not sure if this works but I am also not sure if the mark-as-deleted approach works. Symbolic stores are singly-linked list but they are partially shared among different contexts, at least in my original design. So, if we overwrite or mark entries as deleted in a given symbolic store we may change symbolic stores of other contexts as well, namely contexts that were created earlier. What we need is a merge of symbolic stores that avoids that.

ckirsch: I suggest to neither overwrite nor mark entries as deleted in symbolic stores but instead implement a proper merge.

ChrisEdel: In my version each context has its own symbolic store, so it is no longer shared. Since it may very well be that two contexts overwrite the 'same' variable differently, I think it is even necessary that each context has its own symbolic store. So the approach of overwriting existing variables works for sure and is also more elegant than marking as deleted. However, are you sure that a shared symbolic store is possible or a better solution, respectively?

ckirsch: It may be important to keep algorithmic complexity down. I think, shared stores should work, yes, and may help with scalability. However, if your implementation is correct, then keep it for now. You can come back to this issue later. Then, I’d not copy symbolic stores at all but remember in each context the head pointer to the store whenever the context is branched into two contexts. That pointer marks the shared portion of the store. New stores can then be added to the store by updating existing entries but only in the unshared portion, that is, in the unshared portion there is at most one entry per vaddr. Merging (the unshared portion of) two stores is then easy.

ChrisEdel: Ok, I will take care of the other things first and deal with this when the other things are done.

selfie.c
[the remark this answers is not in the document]
ChrisEdel: Done (for division by zero and invalid concrete memory addresses). However, I noticed that during symbolic execution it is only checked whether the virtual address is a valid address, not if the virtual address is also mapped, which could be the case in `constrain_ld`. I did not modify that since I do not know if there is a specific reason for that. Should I implement that check?

selfie.c
ChrisEdel: We could also just compare the stack pointer, right? At a mergeable location the stack pointer of both contexts must be the same, if they are in different recursion depths, this is not the case.

selfie.c
ChrisEdel: I am a little unsure about this approach. Let us look at the following example: `c1` (context1) encounters a `beq` and creates `c2` , then both set a pointer to the end of the shared store and then insert their new stores into the unshared store. Now both `c1` and `c2` encounter another `beq` and create `c3` (from `c1`) and `c4` (from `c2`). The shared store between `c1` and `c3` is now different than the shared store between `c1` and `c2`, the same is true for the shared store between `c2` and `c4`. That means `c1` and `c2` have now different pointers than before. After (in the best case) `c1` is merged with `c3` (resulting in `c1`) and `c2` with `c4` (resulting in `c2`),  `c1` probably will be merged with `c2`. In that case, you would have to update the pointer again, so that it marks the end of the shared store correctly. So, you have to have at least one stack of these pointers. However, if a context makes an exit before it is merged (e.g. `c3` timeouts before it can be merged with `c1`), the pointer of the corresponding context (in this case `c1`) has to be adapted, in order to know which part of the store is shared and unshared in relation to the context it is going to be merged with. So, you would also need some additional information about how (more specifically by whom) the context was created. Or do I miss something and there is actually an easier way to do that?

selfie.c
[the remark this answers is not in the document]
ckirsch: The reason is that the property of a page to be mapped is complex. It depends on the kernel that does the mapping of pages. So, no, don’t implement that check.

selfie.c
ckirsch: I am not sure if my idea of how to do this is correct. I can only say that each symbolic store is a singly-linked list of vaddr-to-value mappings. All symbolic stores together form a tree of such mappings. The tree structure corresponds to the control flow of the code with the beq instructions corresponding to the nodes in the tree. Whenever we merge any two symbolic contexts they form a `Y` with the lower part of the `Y` being the shared portion and the two upper parts of the `Y` being their respective unshared portions. The only thing to remember is where they split in the `Y`. Then, we need to compute a `|` by keeping the shared portion and merging the two unshared portions.

ckirsch: But I can see that there may be problems merging contexts in arbitrary order, not sure.

selfie.c
ckirsch: Maybe. We need to be sure that this is enough. The stack pointers may be the same by accident but the way we got there may still be different, not sure.

selfie.c
ChrisEdel: Yes, that is the problem, it cannot be guaranteed that all contexts are merged together in the 'right' order (since a context can timeout or exit before reaching its merge location).
But here is another idea: What if the whole symbolic store is shared among all contexts with each entry having an additional value that indicates to which context it belongs to (so the symbolic memory word would have an additional value)? New stores are just added at the top of the symbolic memory, without any overwriting or marking as deleted. So, branching into two contexts (if we copy the pointer mentioned later) and adding new stores would cost us constant time. The only downside is, that the store will grow relatively large. However, each context could have another pointer which always points to his last inserted value. That means, if we search for a value, we do not start looking from the top of the symbolic store, but just from the position the pointer points to. So, we could still take advantage of temporal locality. With this additional pointer, searching for values should have a (in most cases) an acceptable runtime. What do you think about that?

ChrisEdel: Unfortunately, the idea I mentioned before (with the whole memory shared) doesn't work so easily. But I may have found a way to make your idea work. I'll have to formalize it a bit, I'll let you know.

ChrisEdel: I have implemented the (partially) shared symbolic memory according to your idea with some modifications (marking the begin of the shared symbolic memory in the symbolic memory itself instead of using pointers in the contexts). The merging is actually more complicated now, because everything needs to be stored in the right 'place' and that changes with every merge/context. However, merging is also more efficient than before, as are the other operations (inserting, branching, searching). So, the (partially) shared memory does improve the performance by quite a lot!

I think, however, I have to work on the comments so that it becomes clearer what happens in the memory.

PR #148, opened 2019-11-16
Add parameter name to exit() function declaration

Blazefrost:
While compiling a function declaration, selfie expects a parameter to
have both a type as well as an identifier (by calling compile_variable).

As the exit() function declaration's parameter does not have an
identifier, the compilation of assignment `fork-wait` will emit syntax
errors.


Blazefrost:
Thank you for your review, @ckirsch . I've updated the pull request to remove the newline at the end of the file.

PR #151, opened 2019-12-07
Make parents wait for their child within the grader's lock tests

fischer-martin:
If students implemented the `fork-wait` assignment in a way where the entire machine exits as soon as the init-process exits, the outcome of this assignment is dependent on the way they implemented their (probably round-robin) scheduler.
This PR fixes this behaviour by making the parents wait for their children before exiting.


fischer-martin:
As a second thought: It may be better to leave the tests as they are right now since the probability of having an incompatible scheduler is pretty low and this way there isn't a dependency on `wait`.

PR #152, opened 2019-12-11
changes in function replace_extension

Gotsiller:
before, filenames like "../file.c" were changed to ".btor2" instead of "../file.btor2"


selfie.c
```diff
@@ -10024,22 +10024,21 @@ char* replace_extension(char* filename, char* extension) {
...
-
-    if (c == '.') {
-      store_character(s, i, 0);
-
-      c = 0;
-    } else {
-      i = i + 1;
+  // ignore extension
+  while (c != '.'){
+    i = i-1;
```
ckirsch: Please use the convention `i = i - 1;`

PR #153, opened 2020-01-04
Fix race condition in thread assignment

Son-Typ:
Our implementation of `malloc` is not thread safe but the assignment assumes it is. `malloc` might return the same address if several threads call it concurrently.

This PR fixes the race condition by using a shared block of memory. Each thread uses its id as an offset to avoid overwriting other thread's data.

### DIFF
--- a/grader/assignments/thread/syscalls.c
+++ b/grader/assignments/thread/syscalls.c
@@ -16,20 +16,15 @@ uint64_t sumChilds(uint64_t offset, uint64_t pid, uint64_t acc) {
 }
 
 uint64_t get_unique_offset(uint64_t pid1, uint64_t pid2, uint64_t pid3) {
-  uint64_t offset;
+  uint64_t bit0;
+  uint64_t bit1;
+  uint64_t bit2;
 
-  offset = 0;
+  bit0 = pid3 != 0;
+  bit1 = pid2 != 0;
+  bit2 = pid1 != 0;
 
-  if (pid1 != 0)
-    offset = offset + 1;
-  offset = offset * 2;
-  if (pid2 != 0)
-    offset = offset + 1;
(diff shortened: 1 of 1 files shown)


fischer-martin:
You should probably change occurrences of `own_pid` to something like `unique_offset` since `get_own_pid`'s return value depends on internal scheduling decisions (see attached image). Otherwise your fix works perfectly, thanks. :-)

tree

Son-Typ:
Thank you @fischer-martin, I changed the names of the function and variables to better reflect their actual behaviour.
(I remember these trees, I drew one myself for this assignment while debugging this test.)

grader/assignments/thread/syscalls.c
```diff
@@ -2,39 +2,59 @@ void* malloc(unsigned long);
...
+  offset = 0;
+
+  if (pid1 != 0)
+    offset = offset + 1;
+  offset = offset * 2;
+  if (pid2 != 0)
+    offset = offset + 1;
+  offset = offset * 2;
+  if (pid3 != 0)
+    offset = offset + 1;
```
ChristianMoesl: another way to calculate this would be: 
```
((pid1 != 0) * 2 + (pid2 != 0)) * 2 + (pid3 != 0)
```
whereas the self-documenting version would look like: 
```
uint64_t bit0;
uint64_t bit1;
uint64_t bit2;

bit0 = pid3 != 0;
bit1 = pid2 != 0;
bit2 = pid1 != 0;

return (bit2 * 2 + bit1) * 2 + bit0;
```

Son-Typ: Since selfie strives for simplicity I chose the version with the higher readability.

ckirsch:
Nice! Thanks a lot.

PR #156, opened 2020-02-26
Add run on repl.it badge to README

amasad:
Repl.it is an online IDE that's much (much) faster than Cloud9, you also don't need to sign up and the environment comes pre-setup. 

This adds a run on repl.it badge. Your students/users can then just click on that and get started in a matter of seconds. Here is a screenshot of selfie self-compiling

Screen Shot 2020-02-26 at 3 29 12 PM

If you want to try it before you merge: https://repl.it/@amasad/selfie


ckirsch:
@amasad This is awesome! Thanks so much!

PR #160, opened 2020-03-09
Avoid null pointer dereferencing due to errors during initialization

### LINKED ISSUE #158: smalloc: possible array out-of-bound access when out of memory in init_library
When running a code analysis with [Frama-C](https://github.com/Frama-C/Frama-C-snapshot) on selfie, I got the following alarms:

```
file        function                kind           alarm
----------------------------------------------------------------------------------------
selfie.c    two_to_the_power_of     mem_access      \valid_read(power_of_two_table + p)
selfie.c    two_to_the_power_of     initialization  \initialized(power_of_two_table + p)

stack: two_to_the_power_of <- get_bits <- load_character <- 
  print_format1 <- printf1 <- smalloc <- init_library <- main
```

What they mean is that, in case the call to `smalloc` fails (due to lack of memory) *while initializing the library*, because we didn't have time to properly allocate and initialize the global pointer `power_of_two_table`, the error message from `smalloc` itself may trigger a null pointer dereferencing.

In practice, this is extremely unlikely to happen; but in case there is some pedagogical interest in such corner cases, I'm reporting it here. One way to avoid it might be to test that `power_of_two_table` is not null before trying to print anything, for instance.

Please feel free to close this issue if you find it nitpicky.

maroneze:
Fixes #158.

This is an attempt at avoiding the issues mentioned in #158, though the creation of a new global variable displeases me. In principle, one could check that `character_buffer` is no longer null (or `binary_buffer`, since it is the last variable initialized in `init_library`), but it might make odd to check that variable inside `smalloc`, which does not use it directly.

Also, this patch could prevent some error messages from being printed, if the error happens *after* `character_buffer` is initialized but before the end of `init_library`.

If this seems to overly complicate the code for little educational value, feel free to close it.


selfie.c
```diff
@@ -2349,6 +2353,10 @@ void put_character(uint64_t c) {
 
     output_cursor = output_cursor + 1;
   } else {
+    if (!init_done) {
```
ckirsch: @maroneze The `!` operator is not supported in C*. Could you please replace it with `init_done == 0`? Thanks!

ckirsch: Also, please avoid using `{` and `}` if there is only one statement in between. Thanks!

selfie.c
```diff
@@ -2641,7 +2649,9 @@ uint64_t* smalloc(uint64_t size) {
     // any address including null
     return memory;
   else if (memory == (uint64_t*) 0) {
-    printf1("%s: malloc out of memory\n", selfie_name);
+    if (init_done) { // cannot print error message if error during init
```
ckirsch: Same here.

PR #164, opened 2020-03-26
Add files via upload

saraseidl:
Grader Workflow Integration.

Runs Grader on every push to master branch. The test to run has to be included in the commit message between square brackets. Like "Assignment#1 finished [hex-literal]"


ckirsch:
@saraseidl Thanks a lot! This is much appreciated.

PR #167, opened 2020-04-10
Change page-table from a linear table to a tree

fischer-martin:
@Blazefrost and I changed the page-table to use a tree instead of a linear table. This reduces memory consumption in basically every case drastically.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -1072,10 +1072,7 @@ uint64_t NUMBER_OF_LEAF_PTES = 512; // == PAGESIZE / REGISTERSIZE
 
 uint64_t PAGESIZE = 4096; // we use standard 4KB pages
 
-uint64_t PAGE_TABLE_LINEAR = 0; // use a linear page table
 uint64_t PAGE_TABLE_TREE   = 1; // use a two-level tree page table
-
-uint64_t PAGE_TABLE_STRUCTURE = 0; // we use the linear page table
 
 // ------------------------ GLOBAL VARIABLES -----------------------
 
@@ -6770,7 +6767,7 @@ uint64_t leaf_PTE(uint64_t page) {
 uint64_t frame_for_page(uint64_t* parent_table, uint64_t* table, uint64_t page) {
   uint64_t* leaf_pte;
 
-  if (PAGE_TABLE_STRUCTURE == PAGE_TABLE_LINEAR)
+  if (PAGE_TABLE_TREE == 0)
     return (uint64_t) (table + page);
   else {
     leaf_pte = (uint64_t*) load_virtual_memory(parent_table, (uint64_t) (table + root_PTE(page)));
@@ -6785,7 +6782,7 @@ uint64_t frame_for_page(uint64_t* parent_table, uint64_t* table, uint64_t page) 
 uint64_t get_frame_for_page(uint64_t* table, uint64_t page) {
   uint64_t* leaf_pte;
 
-  if (PAGE_TABLE_STRUCTURE == PAGE_TABLE_LINEAR)
+  if (PAGE_TABLE_TREE == 0)
     return *(table + page);
   else {
     leaf_pte = (uint64_t*) *(table + root_PTE(page));
@@ -9160,7 +9157,7 @@ void init_context(uint64_t* context, uint64_t* parent, uint64_t* vctxt) {
 
   // allocate zeroed memory for page table
   // TODO: save and reuse memory for page table
-  if (PAGE_TABLE_STRUCTURE == PAGE_TABLE_LINEAR)
+  if (PAGE_TABLE_TREE == 0)
     set_pt(context, zalloc(VIRTUALMEMORYSIZE / PAGESIZE * REGISTERSIZE));
   else
     set_pt(context, zalloc(VIRTUALMEMORYSIZE / PAGESIZE / NUMBER_OF_LEAF_PTES * REGISTERSIZE));
@@ -9418,7 +9415,7 @@ void map_page(uint64_t* context, uint64_t page, uint64_t frame) {
 
   // assert: 0 <= page < VIRTUALMEMORYSIZE / PAGESIZE
 
-  if (PAGE_TABLE_STRUCTURE == PAGE_TABLE_LINEAR)
+  if (PAGE_TABLE_TREE == 0)
     *(table + page) = frame;
   else {
     root_pte = root_PTE(page);
--- a/selfie.c
+++ b/selfie.c
@@ -1072,6 +1072,11 @@ uint64_t NUMBER_OF_LEAF_PTES = 512; // == PAGESIZE / REGISTERSIZE
 
 uint64_t PAGESIZE = 4096; // we use standard 4KB pages
 
+uint64_t PAGE_TABLE_LINEAR = 0; // use a linear page table
+uint64_t PAGE_TABLE_TREE   = 1; // use a two-level tree page table
+
+uint64_t PAGE_TABLE_STRUCTURE = 0; // we use the linear page table
+
 // ------------------------ GLOBAL VARIABLES -----------------------
 
 uint64_t page_frame_memory = 0; // size of memory for frames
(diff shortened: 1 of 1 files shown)


ckirsch:
How much slower does `make all` get? If it does get significantly slower, you probably need to measure by how much memory access slows down and, separately, by how much context switching slows down, to see if we need to do something about that.

selfie.c
```diff
@@ -9443,15 +9506,36 @@ void map_page(uint64_t* context, uint64_t page, uint64_t frame) {
...
+  }
+
+  return translated_table;
+}
+
 void restore_region(uint64_t* context, uint64_t* table, uint64_t* parent_table, uint64_t lo, uint64_t hi) {
   uint64_t frame;
+  uint64_t* deref_table;
+
+  deref_table = dereference_root_table(parent_table, table);
```
ckirsch, line 9533: Is this really necessary? How about generalizing `is_virtual_address_mapped`?

fischer-martin:
> How much slower does `make all` get? If it does get significantly slower, you probably need to measure by how much memory access slows down and, separately, by how much context switching slows down, to see if we need to do something about that.

comparison of `make all` runtimes:  
make_all

We'll definitely gonna measure the time differences for memory access and context switching between `master` and `tree-pt`.

ckirsch:
Oops. So, that’s the issue really. I wonder if we need to go back to an array then but smaller up to the size of physical memory and then a separate stack page that can be accessed fast. Check if all stack accesses are done relative to the sp register. If yes, you may be able to modify the validity check for virtual addresses to make sure that nothing above the array-mapped memory is accessed other than through the sp register.

Blazefrost:
> Check if all stack accesses are done relative to the sp register.

We looked into the compiler and stack accesses are done relative to REG_SP (for return address and call arguments)  or REG_FP (for local variables).

We also checked how many pages are required by `make all` and it only required a single page, save for `minster` which maps all pages at the beginning of execution.

> I wonder if we need to go back to an array then but smaller up to the size of physical memory and then a separate stack page that can be accessed fast.

We've considered this idea, but we fear that excluding a single stack page from the page table would introduce a lot of changes because it would require to separately handle this case in a lot of functions related to virtual memory. Furthermore, if additional stack pages are required, it would not be trivial to add these.

--------------------
In the recent changeset, we've introduced two optimizations to our PR:

* In commit `d59ccb7`, we've hard-coded the factors for the calculation of the first level index and second level index instead of invoking `left_shift` and `right_shift`. These changes decreased execution time of `make all` from 45.5 minutes to 37.25 minutes.
* In commit `7a28a4c`, we've added the cached virtual context root node into the context structure and perform updates to it only after new pages have been mapped. This further reduced execution time of `make all` down to 15 minutes, which is comparable to the current master with 14.25 minutes.

Blazefrost:
We were successful in getting rid of the `dereference_root_table` function. In the end, the only function that required an additional reference to the hypervisor page table to get hypster running again was `page_for_frame`, which is used in `restore_region` only.

Changing `get_frame_for_page` was not necessary because hypster indirectly passes pages in its own address space to its contexts because `palloc` is page-aligned. This is exploited in `restore_region` where the frame for the virtual context's frame is mapped into the cached context (`selfie.c:9493`)

All changes to `restore_region` were reverted.

Blazefrost:
We also adapted terminology from the RISC-V privileged specification, as described in commit 
22f8ca7.

Furthermore, the calculation for the root page table index and the leaf page table index are not hard-coded anymore, but are based on the count of leaf page table entries.

fischer-martin:
There are two things we aren't sure about though:
* The name `hypervisor_table` for the new parameter in `frame_for_page` (https://github.com/cksystemsteaching/selfie/blob/16841e94c86b9fa279529009bc36472e4fb4c632/selfie.c#L6815). It's different from how that table is named normally (which is `parent_table`) but we thought it would be easier to understand this way.
* The size calculation for the root page-table (https://github.com/cksystemsteaching/selfie/blob/16841e94c86b9fa279529009bc36472e4fb4c632/selfie.c#L9200). We thought about ways to make this expression more readable for quite some time but sadly we couldn't find a good solution.

selfie.c
```diff
@@ -1040,7 +1040,10 @@ void init_memory(uint64_t megabytes);
 uint64_t load_physical_memory(uint64_t* paddr);
 void     store_physical_memory(uint64_t* paddr, uint64_t data);
 
-uint64_t frame_for_page(uint64_t* table, uint64_t page);
+uint64_t root_page_table_index_for_page(uint64_t page);
```
ckirsch, line 1043: In fact, how about just `root_PTE`?

selfie.c
```diff
@@ -1040,7 +1040,10 @@ void init_memory(uint64_t megabytes);
...
+uint64_t leaf_page_table_index_for_page(uint64_t page);
```
ckirsch, line 1044: And `leaf_PTE`?

selfie.c
```diff
@@ -1040,7 +1040,10 @@ void init_memory(uint64_t megabytes);
...
+
+uint64_t frame_for_page(uint64_t* table, uint64_t* hypervisor_table, uint64_t page);
```
ckirsch, line 1046: I'd still prefer `parent_table` over `hypervisor_table`. The parent is the context that created the context of `table`...

selfie.c
```diff
@@ -1067,6 +1070,8 @@ uint64_t WORDSIZEINBITS = 32;
 uint64_t INSTRUCTIONSIZE = 4; // must be the same as WORDSIZE
 uint64_t REGISTERSIZE    = 8; // must be twice of WORDSIZE
 
+uint64_t LEAF_PAGE_TABLE_ENTRIES = 512; // i.e. one leaf page table has the size PAGESIZE
```
ckirsch, line 1073: How about `uint64_t NUMBER_OF_LEAF_PTES = 512; // == PAGESIZE / REGISTERSIZE`?

selfie.c
```diff
@@ -6799,12 +6804,35 @@ void store_physical_memory(uint64_t* paddr, uint64_t data) {
...
-  return (uint64_t) (table + page);
+uint64_t root_page_table_index_for_page(uint64_t page) {
+  return (page / LEAF_PAGE_TABLE_ENTRIES);
+}
+
+uint64_t leaf_page_table_index_for_page(uint64_t page) {
+  return (page - root_page_table_index_for_page(page) * LEAF_PAGE_TABLE_ENTRIES);
+}
+
+uint64_t frame_for_page(uint64_t* table, uint64_t* hypervisor_table, uint64_t page) {
```
ckirsch, line 6815: Please list `parent_table` as first parameter.

selfie.c
```diff
@@ -6799,12 +6804,35 @@ void store_physical_memory(uint64_t* paddr, uint64_t data) {
...
 }
 
 uint64_t get_frame_for_page(uint64_t* table, uint64_t page) {
-  return *(table + page);
+  uint64_t* leaf_pt;
+
+  leaf_pt = (uint64_t*) *(table + root_page_table_index_for_page(page));
+
+  if (leaf_pt == (uint64_t*) 0)
+    // page isn't mapped if the corresponding internal node isn't mapped
```
ckirsch, line 6832: `// page is unmapped if its leaf PTE is unmapped`

selfie.c
```diff
@@ -6799,12 +6804,35 @@ void store_physical_memory(uint64_t* paddr, uint64_t data) {
...
+
+  if (leaf_pt == (uint64_t*) 0)
+    return 0;
+  else
+    return (uint64_t) (leaf_pt + leaf_page_table_index_for_page(page));
 }
 
 uint64_t get_frame_for_page(uint64_t* table, uint64_t page) {
-  return *(table + page);
+  uint64_t* leaf_pt;
```
ckirsch, line 6827: `leaf_pte` rather than `leaf_pt`?

selfie.c
```diff
@@ -9416,12 +9444,24 @@ void save_context(uint64_t* context) {
 
 void map_page(uint64_t* context, uint64_t page, uint64_t frame) {
   uint64_t* table;
+  uint64_t* leaf_pt;
+  uint64_t  root_pt_index;
```
ckirsch, line 9448: `root_pte`?

fischer-martin:
> try to include code that allows going back to a flat page table

Do you mean a command-line argument that specifies which data structure is used?

ckirsch:
> > try to include code that allows going back to a flat page table
> 
> Do you mean a command-line argument that specifies which data structure is used?

No. Just a flag in the source code will do.

selfie.c
```diff
@@ -1065,8 +1068,15 @@ uint64_t WORDSIZEINBITS = 32;
 uint64_t INSTRUCTIONSIZE = 4; // must be the same as WORDSIZE
 uint64_t REGISTERSIZE    = 8; // must be twice of WORDSIZE
 
+uint64_t NUMBER_OF_LEAF_PTES = 512; // == PAGESIZE / REGISTERSIZE
+
 uint64_t PAGESIZE = 4096; // we use standard 4KB pages
 
+uint64_t PAGE_TABLE_LINEAR = 0; // use a linear page table
+uint64_t PAGE_TABLE_TREE   = 1; // use a two-level tree page table
```
ckirsch, line 1076: Just use PAGE_TABLE_TREE as flag.

PR #172, opened 2020-05-04
Add null check in emit_instruction

thomaswulz:
This change allows us to generate machine code using selfie's `emit_*` functions when linking to it as a library.  Otherwise, `emit_instruction` segfaults because `code_line_number` only gets set to a non-null value if selfie is invoked as a compiler.


selfie.c
```diff
@@ -12495,4 +12496,4 @@ int main(int argc, char** argv) {
   init_library();
 
   return selfie();
-}
\ No newline at end of file
+}
```
ckirsch, line 12499: Please remove any unnecessary changes.

PR #173, opened 2020-05-13
Fix behavior of open_write_only on Windows

thomaswulz:
When `open_write_only` is used to open an existing file on Windows, the Mac call goes through, which results in the file not being opened in binary mode. This can be fixed by adding the `_O_BINARY` flag from Windows to the Mac call, while hopefully not breaking anything on Mac.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -5896,17 +5896,16 @@ uint64_t open_write_only(char* name) {
   // not always work and require intervention
   uint64_t fd;
 
-  // try Mac flags
-  // add Windows _O_BINARY flag (32768 = 0x8000) as a hack to make sure existing files are opened properly on Windows
-  fd = sign_extend(open(name, MAC_O_CREAT_TRUNC_WRONLY + 32768, S_IRUSR_IWUSR_IRGRP_IROTH), SYSCALL_BITWIDTH);
+  // try Windows flags
+  fd = sign_extend(open(name, WINDOWS_O_BINARY_CREAT_TRUNC_WRONLY, S_IRUSR_IWUSR_IRGRP_IROTH), SYSCALL_BITWIDTH);
 
   if (signed_less_than(fd, 0)) {
-    // try Linux flags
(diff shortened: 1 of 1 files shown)


selfie.c
```diff
@@ -5897,7 +5897,8 @@ uint64_t open_write_only(char* name) {
   uint64_t fd;
 
   // try Mac flags
-  fd = sign_extend(open(name, MAC_O_CREAT_TRUNC_WRONLY, S_IRUSR_IWUSR_IRGRP_IROTH), SYSCALL_BITWIDTH);
+  // add Windows _O_BINARY flag (32768 = 0x8000) as a hack to make sure existing files are opened properly on Windows
+  fd = sign_extend(open(name, MAC_O_CREAT_TRUNC_WRONLY + 32768, S_IRUSR_IWUSR_IRGRP_IROTH), SYSCALL_BITWIDTH);
```
ckirsch, line 5901: This seems to break the macOS build, see the GitHub workflow for that. Any idea how to do this differently?

thomaswulz:
The new commit, which just tries the original arguments in a different order, seems to work on every platform. But I haven't been able to test if existing files get opened properly on Mac, since I don't have access to one. Perhaps it would make sense to add a new make target for testing the `open` syscall?

ckirsch:
Thanks! What do you mean by testing the open syscall? The thing is that I'd like to avoid introducing platform dependency in the makefile to keep things as simple as possible.

thomaswulz:
I meant adding a target that verifies that `open_write_only` works as it should, i.e. that existing files are overwritten properly, since that's what was broken on Windows. For example, the target could call the selfie compiler so that the generated binary overwrites an existing file, then run `diff` to verify that the file matches the compiler output when writing to a new file. Such a test would've caught the Windows bug.

ckirsch:
Got it. But never mind. It seems to work now.

ckirsch:
@thomaswulz We need to revisit this issue. Opening files on macOS does not work properly anymore. If a file already exists, selfie now appends data to the file rather than overwriting it. Could you please look into this?

PR #181, opened 2020-06-25
fix bug in multi-line comments

AlexLoitzl:
When terminating a multi-line comment with an even number of asterisks we ignore the last asterisk before the terminating "/" and get an error message for "runaway multi-line comment". We need to do a lookahead of one, afer an asterisk to find the end, but then ignore the following character. Therefore the simplest solution providing easiest readability seems to be looping over all asterisks.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -2626,6 +2626,13 @@ uint64_t find_next_character() {
 
         get_character();
 
+        while (character == CHAR_ASTERISK) {
+          // loop over consecutive asterisks to check for "*/"
+          number_of_ignored_characters = number_of_ignored_characters + 1;
+
+          get_character();
+        }
+
         if (character == CHAR_SLASH) {
           // multi-line comments end with "*/"
           in_multi_line_comment = 0;


ckirsch:
@AlexLoitzl Could please post an example here that triggers the bug? Thanks!

AlexLoitzl:
You actually just compile a file containing the following:
`/***/`

ckirsch:
@AlexLoitzl Nice catch! How about changing the outer `if` into a `while`? Does that fix the problem with less code?

AlexLoitzl:
@ckirsch Well, if one would keep the check for the "/" within the loop, we would somehow have to break out of it. Because if an asterisk would follow a multiline comment, we would continue looping. If you do the check after the loop, we don't know if we read an asterisk. One could also store that information in a local variable, but i thought the loop, although having code duplication, was nicer.

ckirsch:
@AlexLoitzl True! How about the code I just committed? https://github.com/cksystemsteaching/selfie/commit/92143a5aaf00f86eb7aef404c869c761c91a77eb

AlexLoitzl:
@ckirsch Yes of course, I actually just wanted to tell you, i actually find a similar solution as well. I guess 8am is too early to review such things. Its, slightly different:
https://github.com/AlexLoitzl/selfie/commit/57c1eb0fa9ad3a69c27b736d9dedec14a7cc3877

ckirsch:
@AlexLoitzl Nice! I took that and polished the overall design. Should be much easier to understand now. Thanks a lot!

PR #183, opened 2020-07-08
Fix: Monster error reporting

ChristianMoesl:
Monster generates invalid SMT-LIB files in error cases. In error scenarios the messages are written into the SMT-LIB files instead of the console. 

image

Also the Makefile is written in a way, were a subsequent error when executing Boolector is silenced. 
A build log from a successful build of the master branch:
image


ckirsch:
> Monster generates invalid SMT-LIB files in error cases. In error scenarios the messages are written into the SMT-LIB files instead of the console.
> 
> image
> 
> Also the Makefile is written in a way, were a subsequent error when executing Boolector is silenced.
> A build log from a successful build of the master branch:
> image

@ChristianMoesl Thank you for fixing this! Does the invalid segment error still occur? If yes, I need to look into it since I introduced segmentation recently which probably causes that problem.

PR #190, opened 2020-08-26
Bare Metal support

Blazefrost:
As part of the bachelor project of @fischer-martin and me, we'd like to introduce bare metal support for selfie. The bare metal support has been implemented in two ways:

- **Library Mode**: selfie.c is linked together with a slim unikernel. System functions are provided by mere function calls
- **Kernel Mode**: selfie self-compiles into a binary, which is then loaded by the kernel as init process. System functions are provided by ecalls (or more generally traps). Paging is required to create virtual address spaces.

All modes provide the interface selfie expects, i.e. `openat`, `read`, `write`, `brk` and `exit` and do actually share the same base code for the I/O operations.


selfie.c
```diff
@@ -5594,8 +5594,9 @@ void store_data(uint64_t baddr, uint64_t data) {
 void emit_instruction(uint64_t instruction) {
   store_instruction(binary_length, instruction);
 
-  if (*(code_line_number + binary_length / INSTRUCTIONSIZE) == 0)
-    *(code_line_number + binary_length / INSTRUCTIONSIZE) = line_number;
+  if (code_line_number != (uint64_t*) 0)
+    if (*(code_line_number + binary_length / INSTRUCTIONSIZE) == 0)
+      *(code_line_number + binary_length / INSTRUCTIONSIZE) = line_number;
```
Blazefrost, line 5599: These changes were introduced by Thomas Wulz before we did the branch split (commit 120e2b71).

We made no changes whatsoever to Selfie. Selfie as currently in the `master` branch will work perfectly fine

PR #194, opened 2020-08-30
Merge branch symbolic into master branch

Sebastian-Landl:
- Updated monster.c
  - Global call stack tree instead of linked lists for each context
  - New schedule, that checks mergeability after every instruction
  - Removed exception handling for `merge` and `recursion` as they are not needed anymore
- Updated selfie.c
  - Removed exceptions `merge` and `recursion` as they are not needed anymore
- Add bachelor_thesis_landl.pdf


tools/monster.c
```diff
@@ -1349,141 +1276,42 @@ char* smt_ternary(char* opt, char* op1, char* op2, char* op3) {
...
+  if (merge_enabled == 0)
     return;
-  }
 
   print("; merging two contexts at ");
   print_code_context_for_instruction(location);
 
   if (debug_merge)
-    printf2(" -> active context: %u, mergeable context: %u", (char*) active_context, (char*) mergeable_context);
+    printf2(" -> active context: %d, mergeable context: %d", (char*) active_context, (char*) mergeable_context);
```
ckirsch, line 1314: This should remain %u, no?

tools/monster.c
```diff
@@ -2096,12 +1830,13 @@ void monster(uint64_t* to_context) {
...
 
-  timeout = max_execution_depth - get_execution_depth(to_context);
+  timeout = 1;
 
   while (1) {
+
     if (debug_merge)
       if (from_context != (uint64_t*) 0)
-        printf4("; switching from context %u to context %u (merge locations: %x, %x)\n",
+        printf4("; switching from context %d to context %d (merge locations: %x, %x)\n",
```
ckirsch, line 1839: Same as above.

tools/monster.c
```diff
@@ -216,29 +215,25 @@ char* smt_unary(char* opt, char* op);
...
+
+void      add_child(uint64_t* parent, uint64_t* child);
+void      step_into_call(uint64_t* context, uint64_t address);
+void      step_out_of_call(uint64_t* context);
 
 void use_stdout();
 void use_file();
 
-void monster(uint64_t* to_context);
+uint64_t monster(uint64_t* to_context);
```
ckirsch, line 236: Should stay void, no?

tools/monster.c
```diff
@@ -2076,11 +1808,13 @@ void use_file() {
   output_fd   = smt_fd;
 }
 
-void monster(uint64_t* to_context) {
+uint64_t monster(uint64_t* to_context) {
```
ckirsch, line 1811: Same as above.

ckirsch (changes requested):
This should say MIPSTER in comment:

https://github.com/cksystemsteaching/selfie/blob/b4eb52571c90f3c90a88a8492d7eafd7789ba767/tools/monster.c#L541

tools/monster.c
Sebastian-Landl: I must have overlooked that change. I actually think both would work fine, because it's just about identifying contexts for debugging. I changed it to %u.

tools/monster.c
Sebastian-Landl: Overlooked that as well just like before, also changed it to %u.

tools/monster.c
Sebastian-Landl: I overlooked that too, fixed as well.

tools/monster.c
Sebastian-Landl: Fixed it here too.

tools/monster.c
```diff
@@ -155,21 +155,19 @@ uint64_t* copy_symbolic_context(uint64_t* original, uint64_t location, char* con
...
 }
 
 uint64_t  get_execution_depth(uint64_t* context) { return             *(context + 19); }
 char*     get_path_condition(uint64_t* context)  { return (char*)     *(context + 20); }
 uint64_t* get_symbolic_memory(uint64_t* context) { return (uint64_t*) *(context + 21); }
 uint64_t* get_symbolic_regs(uint64_t* context)   { return (uint64_t*) *(context + 22); }
 uint64_t  get_beq_counter(uint64_t* context)     { return             *(context + 23); }
-uint64_t  get_merge_location(uint64_t* context)  { return             *(context + 24); }
 uint64_t* get_merge_partner(uint64_t* context)   { return (uint64_t*) *(context + 25); }
 uint64_t* get_call_stack(uint64_t* context)      { return (uint64_t*) *(context + 26); }
```
ckirsch, line 172: Here's a bug in the offset now.

tools/monster.c
```diff
@@ -178,7 +176,6 @@ void set_path_condition(uint64_t* context, char* path)        { *(context + 20)
 void set_symbolic_memory(uint64_t* context, uint64_t* memory) { *(context + 21) = (uint64_t) memory; }
 void set_symbolic_regs(uint64_t* context, uint64_t* regs)     { *(context + 22) = (uint64_t) regs; }
 void set_beq_counter(uint64_t* context, uint64_t counter)     { *(context + 23) =            counter; }
-void set_merge_location(uint64_t* context, uint64_t location) { *(context + 24) =            location; }
```
ckirsch, line 182: Same as above.

ckirsch (changes requested):

tools/monster.c
Sebastian-Landl: Thanks, it's fixed.

tools/monster.c
Sebastian-Landl: Fixed as well

ckirsch (approved):

PR #198, opened 2020-09-16
Descriptions for Grader assignments (fixes #189)

### LINKED ISSUE #189: Add a reference to assignment descriptions to the grader
@aemonk summarized all assignment descriptions from the cc2020 slack channel into a Markdown file as a quick reference. In the grader, we could add references back to the descriptions of each assignment so that the aim of each assignment is easier to understand.

Reference: https://cksystemsteaching.slack.com/archives/CUQUKPD61/p1598390644002700

Blazefrost:
This PR introduces a `description` field to the `Assignment` class and furthermore assigns links to the descriptions.

@aemonk created a nice reference for all assignments on the `grader` branch. This PR will merge parts of the `grader` branch up until the last commit that introduced changes to the reference.

The links will result in an 404 error until the branch has been merged because it references the `master` branch.


Blazefrost:
Right now, the description value is not used, though. It is solely useful for people that inspect the source. I tried to print them after the assignment name, but I think it looks rather messy:
image

Maybe we could introduce a flag to print the URI of a given assignment?

aemonk:
What if we add `/grader/compiler-assignments.md` to the heading "Compiler Assignments" in the self.py output?  
Then the students could look for themselves and since we have equality called assignments, this should be easy to do, right?

ckirsch:
Can you reorder how the assignments appear in the output of `self.py` without any arguments?

Also, please rename "Usage:" to "synopsis:".

ckirsch:
The assignments should appear in the order of how they are assigned in class.

PR #217, opened 2020-10-01
Fix compiler warnings when compiling with GCC 10.2.0 (fixes #216)

### LINKED ISSUE #216: Compiler warnings with GCC 10.2.0
Compiling the current version of selfie (commit dcde9b6) with GCC version 10.2.0 causes compiler warnings. This should be fixed in order to not cause problems with the autograder.

```selfie.c: In function ‘print_gc_profile’:
selfie.c:7723:3: warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]
 7723 |   if (is_gc_library(context) == 0) print(" (external)"); println();
      |   ^~
selfie.c:7723:58: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the ‘if’
 7723 |   if (is_gc_library(context) == 0) print(" (external)"); println();
      |                                                          ^~~~~~~
selfie.c: In function ‘selfie_compile’:
selfie.c:7193:38: warning: ‘fetch_dss_code_location’ may be used uninitialized in this function [-Wmaybe-uninitialized]
 7193 |   ic_addi = ic_addi - (binary_length - fetch_dss_code_location) / INSTRUCTIONSIZE;
      |                       ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
selfie.c:5222:12: note: ‘fetch_dss_code_location’ was declared here
 5222 |   uint64_t fetch_dss_code_location;
      |            ^~~~~~~~~~~~~~~~~~~~~~~```

fischer-martin:
The `-Wmaybe-uninitialized` is really strange in my opinion. It only occurs when optimization is activated, which is also mentioned on the gcc manpage. There, it also says that sometimes these warnings can happen because the compiler could not verify a code snippet's correctness. What I don't really understand is why the compiler does not see that both branch conditions in `selfie_compile()` depend on `GC_ON` which is never modfied between both of them.

The fix (commit 49db4c9) is a little bit ugly in my opinion but I did not find a better solution. An alternative would be to pass `-Wno-maybe-uninitialized` as a C-flag but I guess this would do more harm than good.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -5221,6 +5221,8 @@ void selfie_compile() {
   uint64_t number_of_source_files;
   uint64_t fetch_dss_code_location;
 
+  fetch_dss_code_location = 0;
+
   // link until next console option
   link = 1;
 
--- a/selfie.c
+++ b/selfie.c
@@ -7720,7 +7720,9 @@ void print_gc_profile(uint64_t* context) {
     (char*) fixed_point_ratio(gc_mem_metadata, MEGABYTE, 2),
     (char*) fixed_point_percentage(fixed_point_ratio(gc_mem_mallocated, gc_mem_metadata, 4), 4),
     (char*) gc_num_ungced_mallocs);
-  if (is_gc_library(context) == 0) print(" (external)"); println();
+  if (is_gc_library(context) == 0)
+    print(" (external)");
+  println();
 }
 
 void gc_arguments() {


ckirsch:
@fischer-martin Your fixes are fine. Never mind the compiler. Proving that `GC_ON` does not change in between is very difficult since it is a global variable.

PR #221, opened 2020-10-17
grader: updated tests for assembler-parser

FrieAT:
This PR addresses issues to the grader test `assembler-parser`:

- dollar sign ($)is not used anymore within `./selfie -c selfie.c -s selfie.s` for registers
- removed deprecated test `missing-address.s`
- added other 2 type of tests for testing all registers and for testing an example for an invalid parameter


ckirsch:
@FrieAT Awesome! Thanks a lot!

PR #225, opened 2020-10-21
Machine: Clean opensbi platform directory after selfie-opensbi.elf build

Blazefrost:
OpenSBI's makefile target `all` declares prerequisites on the
fw_payload.bin file of the specified platform, which itself depends on
the fw_payload.elf file that we move out after the build has finished.
As the fw_payload.bin file still exists, make does not look further and
does not rebuild fw_payload.elf.

Instead, remove the whole firmware/ subdirectory of the platform's build
output directory. This is preferrable to a `make clean` because this
would also clean shared build artifacts.

The relevant lines of this behavior are (in OpenSBI v0.7):
- opensbi/firmware/objects.mk:30
- opensbi/Makefile:134
- opensbi/Makefile:300
- opensbi/Makefile:304


ckirsch:
@Blazefrost Do you know why this became a problem now and not earlier?

Blazefrost:
`Dockerselfie` sets environmental variable `MAKEFLAGS=-j4`, enabling Make to perform the build using 4 tasks. I think this may have resulted in a race condition because our Makefile didn't clean OpenSBI's build directory sufficiently. 

Now that I think of it, I fear there is another PR necessary to fix it for good as the way the OpenSBI images are built right now won't work with concurrent tasks. I will have to isolate the `kernel` and `library` OpenSBI builds into separate build output directories at the cost of a little longer build time, but it won't change anything about our own build output directory.

Issue #229, opened 2020-11-04
`self-assembler` assignment results in timeout


fischer-martin:
It has been reported to me that the the [current test](https://github.com/cksystemsteaching/selfie/blob/master/grader/self.py#L212) for the `self-assembler` assignment causes a timeout (runtime of 3-5min vs. an allowed maximum of 60s). Since increasing the timeout isn't practical (especially since the bulk-grader doesn't run in parallel), I suggest that we change the test a little bit.

fischer-martin:
Thoughts about that, @ckirsch @Blazefrost ?

ckirsch:
How would you change the test?

fischer-martin:
Maybe compile selfie once from the C* source and assemble it from the output generated by `-s` and then compare the resulting binaries.

ckirsch:
I thought that is what self-assembler checks...what does it actually check?

fischer-martin:
See for yourself:

```def check_self_assemblation() -> List[Check]:
    return check_execution('./selfie -c selfie.c -s selfie1.s -a selfie1.s -m 128 -a selfie1.s -s selfie2.s ',
                           'selfie can assemble its own binary file') + \
        check_execution('diff -q selfie1.s selfie2.s',
                        'both assembly files are exactly the same')```

Blazefrost:
It seems like both assembling and disassembling takes significantly longer on mipster than just compilation.
Maybe we could:
1. Compile, disassemble, assemble, and store a selfie binary
2. Load the binary into a mipster instance and let it compile and store a selfie binary.
3. Compare the resulting binaries from step 1 and 2

If the assemble step is working correctly, the binary from step 1 should be logically equivalent to the host's selfie and produce the same binary in step 2.

ckirsch:
@Blazefrost Sounds good! Please do a PR.

PR #232, opened 2020-11-13
Grader: Fix assignment `self-assembler` passing without implementation

Blazefrost:
@fischer-martin informed me that the change made by #229 introduced a
bug.
The assignment will falsely succeed when support for the `-a` flag
has been implemented, but no changes to global `binary` have
been made. This is due to the previously compiled and disassembled
binary still being in memory.

This PR splits the first execution check into a preparation step that
compiles and disassembles `selfie.c` and another check that does the
assemble step, binary output, mipster compilation and binary output.

### DIFF
--- a/grader/self.py
+++ b/grader/self.py
@@ -213,9 +213,9 @@ def check_self_assemblation() -> List[Check]:
     return check_execution('./selfie -c selfie.c -s selfie1.s',
                            'preparation: selfie can compile and assemble its own source', mandatory=True) + \
         check_execution('./selfie -a selfie1.s -o selfie1.m -m 128 -c selfie.c -o selfie2.m',
-                        'selfie can re-assemble its own binary file') + \
+                        'selfie can re-assemble its own binary file', mandatory=True) + \
         check_execution('diff -q selfie1.m selfie2.m',
-                        'both binary files are exactly the same')
+                        'both binary files are exactly the same', mandatory=True)
 
 
 def check_processes() -> List[Check]:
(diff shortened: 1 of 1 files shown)


Blazefrost:
I tested it with this minimal patch to selfie. It will be accepted by `master`, but not by this PR fix.

``` diff
diff --git a/selfie.c b/selfie.c
index f2ddabc..7c59954 100644
--- a/selfie.c
+++ b/selfie.c
@@ -10223,6 +10223,8 @@ uint64_t selfie(uint64_t extras) {
         selfie_disassemble(1);
       else if (string_compare(argument, "-l"))
         selfie_load();
+      else if (string_compare(argument, "-a"))
+        get_argument();
       else if (extras == 0) {
         if (string_compare(argument, "-m"))
           return selfie_run(MIPSTER);
```

Blazefrost:
@fischer-martin , could you test whether it still works with correct implementations?

Also, @ckirsch , should we set all `self-assembler` checks to be mandatory? After all, if any of these checks do not work, the whole self assemble step shouldn't be successful, too. If so, I could push another commit to this PR.

fischer-martin:
I've tested some correct solutions with this fix and it looks fine. That'll definitely help students to spot errors within their solutions.

ckirsch:
> @fischer-martin , could you test whether it still works with correct implementations?
> 
> Also, @ckirsch , should we set all `self-assembler` checks to be mandatory? After all, if any of these checks do not work, the whole self assemble step shouldn't be successful, too. If so, I could push another commit to this PR.

Yes, please.

Blazefrost:
Done!

PR #234, opened 2020-11-19
Describe code and data segment in ELF header seperately

ChristianMoesl:


ckirsch:
@ChristianMoesl Awesome! Thanks a lot! There is just a segfault in the 32-bit build. Can you check?

ChristianMoesl:
@ckirsch I fixed it. I somehow thought, the ELF32 and ELF64 program headers are exactly the same. Turns out, they are not :)

ckirsch:
@ChristianMoesl still failing with a segfault.

ChristianMoesl:
@ckirsch Well that was kind of painful to debug. There is a crazy assumption in the x86-translater. The translator only modifies the existing ELF header in memory, as opposed to just create one on it's own. That was probably not a good idea.

Took me hours to debug, because I though there is something wrong in the x86-translation process itself. (Since there are  now padding bytes between the code and the data segment in `binary`).

ckirsch:
@ChristianMoesl My apologies. I forgot that there is this x86 translator which is unfortunately a bit of a hack. It also just occurred to me that there are dependencies in modeler but I will take care of them. I will also polish your code, probably tomorrow, and maybe inline the `encode_elf_program_header` procedure for better readability, despite code duplication.

ChristianMoesl:
@ckirsch  Perfect, thanks! Didn't mean to be rude. 😅

PR #236, opened 2020-11-21
Grader: Allow SIGINT to be handled when waiting for a subprocess (fixes #144)

### LINKED ISSUE #144: Ctrl-C
Pressing Ctrl-C just once should terminate grader.

Blazefrost:
When forking a child process, both the parent and the children reside in
the same process group. Signals like SIGINT that directed to a process
group will be distributed to all processes in the group. However, the
grader did not receive SIGINT when issuing [Ctrl]+[C] in a hosting
terminal. This is due to to the grader using `system` to start the child
processes.

As per the man-pages of `system` for
[FreeBSD](https://www.unix.com/man-page/freebsd/3/system/),
[macOS](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/system.3.html)
and [Linux](https://linux.die.net/man/3/system), signal `SIGINT` will be
ignored while the application waits for the child process to finish.
Unfortunately, the return value of Python's `os.system` is
platform-dependent.

This commit instead uses `subprocess.run` which under the hood calls an
`execve`-like function. SIGINT will not be masked by this and will be
received by both the child `git` processes and the parent grader
process.

### DIFF
--- a/grader/lib/cli.py
+++ b/grader/lib/cli.py
@@ -4,7 +4,7 @@ import sys
 from argparse import ArgumentParser, RawDescriptionHelpFormatter
 from io import StringIO
 from pathlib import Path
-from subprocess import run
+from subprocess import run, DEVNULL
 from typing import Callable, Optional, List, Tuple, Dict, Set
 
 from lib.functional import flatmap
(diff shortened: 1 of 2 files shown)


ckirsch:
@Blazefrost Very nice! Thank you. Could you please check the failing grader target? I am not sure if it is related to your PR.

Blazefrost:
I finally found out what was wrong. It was actually the unit test itself that failed because it mocked a replaced function. See b0f8df3 for a detailed write-up

ckirsch:
@Blazefrost Sounds complicated :/ the build is still failing :/ can you please check?

Blazefrost:
Didn't the checks suceed? While the first commit, 6c66b80, is marked as failed, GitHub says that 9 checks were successful and two are _expected_

ckirsch:
@Blazefrost You are right. Seems to work :)

PR #241, opened 2020-12-10
Grader: Add more checks to `fork-wait-exit`

Blazefrost:
This commit introduces two more checks to the `fork-wait-exit` assignment.

* The first check asserts whether the `wait` implementation handles unmapped virtual addresses and maps pages accordingly. If the syscall does not handle it properly, Selfie will probably raise a SEGFAULT signal.
* The second check asserts whether `wait` handles invalid addresses outside the 4GiB correctly and returns -1 instead of a terminated child's PID.

### DIFF
--- a/grader/assignments/fork-wait/invalid-address.c
+++ b/grader/assignments/fork-wait/invalid-address.c
@@ -11,9 +11,9 @@ uint64_t wexitstatus(uint64_t status) {
 int main() {
   uint64_t sum;
   uint64_t pid;
-  uint64_t* code;
+  uint64_t* heap_top;
 
-  code = malloc(8);
+  heap_top = malloc(8);
   sum = 38;
 
   pid = fork();
@@ -26,8 +26,8 @@ int main() {
 
     // Second invalid access - free region between stack and heap
     // Bump pointer allocation -> we can derive the end of the heap
-    // using the most recently allocated memory block, in this case `code`
-    pid = wait(code + 1);
+    // using the most recently allocated memory block, in this case `heap_top`
+    pid = wait(heap_top + 1);
     sum = sum + pid;
 
     // Third invalid address - outside the 4GiB virtual addressing space
@@ -37,8 +37,8 @@ int main() {
 
   // Actually query the PID/status
   // The previous calls should have failed before and mustn't have consumed the zombie
-  pid = wait(code);
+  pid = wait(heap_top);
 
   // 38 (initial) + 5 (exit code) + 2 (PID) + 3 * (-1) (wait error on invalid vaddr)
-  return sum + pid + wexitstatus(*code);
+  return sum + pid + wexitstatus(*heap_top);
 }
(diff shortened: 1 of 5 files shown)


Blazefrost:
There is a check for illegal addresses still pending if there is enough time left before the assignment is published. 

Also, when an invalid pointer is passed to `wait`, a zombie's status should not be consumed, right? I could extend the second check to assert this behavior.

@fischer-martin could you check it out?

ckirsch:
> There is a check for illegal addresses still pending if there is enough time left before the assignment is published.
> 
> Also, when an invalid pointer is passed to `wait`, a zombie's status should not be consumed, right? I could extend the second check to assert this behavior.
> 
> @fischer-martin could you check it out?

Sounds all good to me. @Blazefrost Is the PR ready or are you still working on it?

Blazefrost:
I would like to extend the invalid pointer check as aforementioned and maybe add a check whether `wait` can handle NULL pointer and just return the PID in this case, if that's okay with you.

ckirsch:
> I would like to extend the invalid pointer check as aforementioned and maybe add a check whether `wait` can handle NULL pointer and just return the PID in this case, if that's okay with you.

Sure! Please go ahead.

Blazefrost:
Done! However, I probably need to check whether an implementation maps a page to NULL. I thought about implementing it by trying to access NULL and provoking a segfault.

@fischer-martin what do you think about the current state of the PR?

grader/assignments/fork-wait/invalid-address.c
```diff
@@ -0,0 +1,44 @@
...
+uint64_t wexitstatus(uint64_t status) {
+  return status * 281474976710656 / 72057594037927936;
+}
+
+int main() {
+  uint64_t sum;
+  uint64_t pid;
+  uint64_t* code;
+
+  code = malloc(8);
```
fischer-martin, line 16: The variable name is misleading. Something like `heap_top` would probably fit better.

grader/assignments/fork-wait/null-ptr.c
```diff
@@ -0,0 +1,15 @@
+uint64_t* NULL = (uint64_t*) 0;
+
+int main() {
+  uint64_t pid;
+  uint64_t i;
+
+  pid = fork();
+  if (pid == 0)
+    exit(0);
+  else {
```
fischer-martin, line 10: The curlies can be removed.

grader/assignments/fork-wait/unmapped-page-wait.c
```diff
@@ -0,0 +1,28 @@
...
+
+  // Enforce an unmapped virtual address by allocating a page-wide array
+  // and not touching it
+  // It must be page-wide to move the last element across potential already
+  // mapped page boundaries
+  // wait uses the last element, then
+  code = malloc(8 * 4096);
+
+  pid = fork();
+  if (pid == 0) {
```
fischer-martin, line 19: The curlies can be removed (same for `else`).

fischer-martin:
@Blazefrost Looks really good! I've only added some annotations regarding cosmetic changes.

ckirsch:
@Blazefrost The travis build still fails though in the grader target...

Blazefrost:
I fixed the redundant braces and fixed the unit test failure. The failure was caused by a check whether the expected exit code overlaps with one of Selfie's exit codes. I also adapted the checks so that they do no longer assume the child PID is necessarily 2.

The additional checks should be done now!

PR #242, opened 2020-12-20
Grader: Add `--truncate` option

Blazefrost:
@fischer-martin asked me to add a truncate option to the grader because long debugging output make grading harder.

The option `--truncate <head> <tail>` takes two integral parameters that specify the amount of lines at the beginning or the end to print, respectively. By passing `-1` to at least one of these parameters, the output is unlimited, i.e. the flag is not effective.

### DIFF
--- a/grader/lib/cli.py
+++ b/grader/lib/cli.py
@@ -1,7 +1,7 @@
 import os
 import re
 import sys
-from argparse import ArgumentParser, RawDescriptionHelpFormatter
+from argparse import ArgumentParser, RawDescriptionHelpFormatter, ArgumentTypeError
 from io import StringIO
 from pathlib import Path
 from subprocess import run, DEVNULL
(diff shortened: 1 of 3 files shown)


Blazefrost:
One thing that still needs debugging is that `argparse` does not complain about missing required arguments even though no assignment name has been passed.

grader/lib/cli.py
```diff
@@ -66,6 +67,17 @@ def parse_assignment(args: str, assignments: Set[Assignment]) -> Optional[Assign
...
+        if value < 0:
+            raise ArgumentTypeError("truncate line count must be a natural number")
+
+        return value
+    except ValueError:
+        raise ArgumentTypeError("truncate line count must be an integer")
+
 
 def validate_options_for(assignment: Optional[Assignment]):
     if not bulk_grade_mode and is_in_quiet_mode() and assignment is None:
```
Blazefrost, line 83: It seems like this is the culprit. Interestingly, it only emits an error that no assignment has been specified if the grader is not in bulk-grading mode (that's fine) _and it is in quiet mode_.
Interestingly, function `error` does not even print the message loudly (i.e. ignoring quiet), but rather silently exits.
It looks like the call to `is_in_quiet_mode()` is not even necessary. This is the only position where the quiet mode state is queried and thus entering it does not have any effect on whether the assignment name is required or not.

ckirsch (approved):
@Blazefrost Nice work! Thanks a lot!

PR #243, opened 2020-12-22
Grader: Add release after exit check for `lock` assignment

Blazefrost:
This PR introduces an additional check to the `lock` assignment.
It asserts whether a lock held by terminated processes are released. The check creates a fork that takes a lock, prints a message and terminates. The parent waits for the child to terminate, tries to take the lock itself, prints a message, unlocks the lock and terminates itself.

### DIFF
--- a/grader/self.py
+++ b/grader/self.py
@@ -266,7 +266,7 @@ def check_lock() -> List[Check]:
                         success_criteria='Hello World!    ' * 8) + \
         check_execution('./selfie -c <assignment>release-after-exit.c -m 128',
                         'Lock is granted to a process after a terminated process did not release its lock',
-                        success_criteria='Hello child!    Hello parent!   ')
+                        success_criteria='Hello child!    Hello parent!   ', timeout=5)
 
 
 def check_threads() -> List[Check]:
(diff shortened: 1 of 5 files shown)


grader/self.py
```diff
@@ -263,7 +263,10 @@ def check_lock() -> List[Check]:
                         success_criteria='Hello World!    ' * 8) + \
         check_execution('./selfie -c selfie.c -m 128 -c <assignment>print-with-lock.c -y 10',
                         '16 processes are printing in sequential order with the use of locks on HYPSTER',
-                        success_criteria='Hello World!    ' * 8)
+                        success_criteria='Hello World!    ' * 8) + \
+        check_execution('./selfie -c <assignment>release-after-exit.c -m 128',
```
fischer-martin, line 267: Is it easily possible to set a separate timeout for this test? (So that it doesn't take too long in case a student didn't implement the implicit unlock upon exit.)

fischer-martin (changes requested):

grader/self.py
Blazefrost: I now explicitly set the timeout to 5 seconds. This should suffice as the check's source is rather small.

fischer-martin (approved):

PR #248, opened 2021-01-23
Support for L1-cache emulation

fischer-martin:
This PR introduces mipster support for a VIPT n-way set-associative write-through L1-cache. It is available through the `-L1` command line argument (e.g. `./selfie -c examples/hello-world.c -L1 128` will compile and run hello-world.c on mipster with 128 MiB memory and an L1 cache).
The terminology has been taken from the RISC-V edition of "Computer Organization and Design" by Patterson and Hennessy.


selfie.c
```diff
@@ -1150,6 +1150,144 @@ uint64_t debug_switch = 0;
...
+
+// ------------------------ GLOBAL CONSTANTS -----------------------
+
+// indicates whether the machine has a cache or not
+uint64_t L1_CACHE_ENABLED = 0;
+
+// machine-enforced coherency for self-modifying code (selfie also
+// works if this is turned off since there is no code modification
+// during runtime and stores in the code segment are illegal)
+uint64_t L1_CACHE_COHERENCY = 0;
```
fischer-martin, line 1250: While the coherency mechanism per se works, it is necessary to allow writes to the code segment by changing `is_valid_write_segment()` in order to actually make use of it.

fischer-martin:
We are also currently working on some examples. These include demonstrations of how different data access patterns affect cache hits/misses and self-modifying code (requires `is_valid_segment_write()` to be modified) which has different behavior depending on if the coherency mechanism between data cache and instruction cache is turned on or off.

ckirsch:
@fischer-martin @mattpaulitsch Really nice work! I will pull this and then do some further polishing work. Please follow that. Thanks!

selfie.c
```diff
@@ -1150,6 +1150,144 @@ uint64_t debug_switch = 0;
...
+
+// example configurations:
+// +-------------------+---------------+-----------------------------+------------+
+// |              name |    cache size |               associativity | block size |
+// +===================+===============+=============================+============+
+// |       CORE-V CVA6 | dcache: 32 KB |                   dcache: 8 |       16 B |
+// |                   | icache: 16 KB |                   icache: 4 |            |
+// +-------------------+---------------+-----------------------------+------------+
+// |     direct-mapped |         2^x B |                           1 |      2^y B |
```
ckirsch, line 1259: What is `x` and `y` actually?

fischer-martin: Those are just some arbitrary exponents (i.e. `^` does not represent the bitwise XOR operator here) since we wanted to express that those values are always powers of 2.

selfie.c
```diff
@@ -9393,6 +9891,16 @@ void print_profile(uint64_t* context) {
     print_register_memory_profile();
   }
 
+  if (L1_CACHE_ENABLED) {
+    print_cache_statistic(get_cache_hits(L1_DCACHE), get_cache_misses(L1_DCACHE), "d");
+    println();
+
+    print_cache_statistic(get_cache_hits(L1_ICACHE), get_cache_misses(L1_ICACHE), "i");
```
ckirsch, line 9898: This reports 1 more fetch than the total number of executed instructions. Where's the bug? :)

fischer-martin: We found the reason for that (and why the statistics aren't "wrong" for all programs): If a load/store causes a page fault the page will be mapped and the same instruction will be executed again (this time successfully). This means that the instruction is fetched two times but `ic_load`/`ic_store` is only incremented one time.
In my opinion both, the number of icache accesses and the number of (successfully) executed instructions, are correct.

ckirsch: Sounds good :)

ckirsch: Check out my changes and please look for errors. It was a bit more work than I anticipated.

fischer-martin: The changes look good. I didn't spot anything that looks wrong.

mattpaulitsch: I don't see any problems with the changes either.

PR #250, opened 2021-01-25
Support for printf

Son-Typ:
This PR introduces support for variadic functions by implementing the macros `var_start`, `var_arg`, and `var_end`; adaptions of the standard `va_start`, `va_arg`, and `var_end` macros. Those macros were used to implement `printf`, `sprintf`, and `dprintf`, which can be used at bootlevel >= 1. For boot level 0, the `printf`, `sprintf`, and `dprintf` implementation by the bootstrapping compiler is used.
The tools (`modeler.c`, `monster.c`, etc.) were also adapted to work with the new functions.
Support for Windows and compilation for bare-metal RISC-V were dropped as was discussed.


Son-Typ:
Regarding the failed check (`Run Docker Selfie`): We have tried to replicate the error without success. We tried:

- macOS (also tested as part of Github Actions)
- Debian 10
- Ubuntu 20.04 (also tested as part of Github Actions)
- Compiling inside the provided Docker image `cksystemsteaching/selfie`
- Building the Docker image ourselves and compiling inside it

The failing `vm` target was successful on all of those systems. Sadly, we do not know if there is a bug and - if there is - how we should find it.

Son-Typ:
Thanks to @Blazefrost we have identified the problem to be yet another call to `make everything` in another file that was interleaved in the error report and therefore hard to find. We have replaced it with `make all` as `machine` is currently not compatible with our implementation. Another problem seems to be a timeout for `btormc` for recursive-ackerman. I don't think this is our fault. While changing the call stacks also changes the model, the results are (obviously) still the same.

PR #257, opened 2021-02-18
Fix code and data segment alignment in binary file

ChristianMoesl:
This patch will let Selfie generate valid ELF binaries for qemu/spike and Monster. Both segments had to be aligned by `p_align` in the binary file. The documentation of that is not really good. It is mentioned [here in the description of p_align](https://man7.org/linux/man-pages/man5/elf.5.html).

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -1007,7 +1007,7 @@ uint64_t e_flags     = 0;  // ignored
 uint64_t e_ehsize    = 64; // elf header size 64 bytes (ELFCLASS64) or 52 bytes (ELFCLASS32)
 uint64_t e_phentsize = 56; // size of program header entry 56 bytes (ELFCLASS64) or 32 bytes (ELFCLASS32)
 
-uint64_t e_phnum = 1; // number of program header entries (code and data segment; TODO: extend to 2)
+uint64_t e_phnum = 2; // number of program header entries (code and data segment)
 
 uint64_t e_shentsize = 0; // size of section header entry
 uint64_t e_shnum     = 0; // number of section header entries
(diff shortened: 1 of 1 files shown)


selfie.c
```diff
@@ -5488,6 +5487,7 @@ void emit_bootstrapping() {
     // avoid sign extension that would result in an additional sub instruction
     if (gp_value < two_to_the_power_of(31) - two_to_the_power_of(11))
       // assert: generates no more than two instructions
+      // assert: generates no data in data segment
```
ckirsch, line 5490: Please explain this comment. I am not sure I understand.

selfie.c
```diff
@@ -6628,10 +6627,14 @@ void selfie_output(char* filename) {
     exit(EXITCODE_IOERROR);
   }
 
+  code_size_with_padding = round_up(code_size, PAGESIZE);
```
ckirsch, line 6630: This should be `p_align`, not `PAGESIZE`, right?

ckirsch:
Please double check that monster.c and modeler.c are ok with these changes.

selfie.c
ChristianMoesl: I documented a bit more verbose in the updated commit.

Essentially the problem comes from the fact, that `load_integer()` does allocate memory on the data segment for big values. But at this point, there are no data segment allocations allowed, since the `gp_value` is already calculated. 

Even tough we do check for `gp_value` here, it is not clear, that we do that for 2 reasons:
 - maximum of `n` amount of emitted instructions in `load_integer()`
 - `load_integer()` does not allocate memory in the data segment

ChristianMoesl:
> Please double check that monster.c and modeler.c are ok with these changes.

I'll have another look later and will give you feedback here.

selfie.c
ChristianMoesl: I also thought of directly emitting the necessary instructions in place, instead of calling `load_integer()`, because it relies on some implementation details of `load_integer()`, which is probably not good.

ChristianMoesl:
I double checked it. All the tools should be fine with this change. In essence it really only affects the binary file format and therefore also `selfie_output()` (write extra padding bytes) and `selfie_load()` (ignore padding bytes).

ckirsch:
Looks great! I only wish I would remember why this failed a while ago when I tried essentially the same thing.

PR #258, opened 2021-02-19
Add Bachelor's Thesis "RISC-V Bare-Metal Library Operating System for Selfie"

Blazefrost:


Blazefrost:
The links to the PDF and release commit won't work right now, because I have to create them after the changes have been merged.

PR #259, opened 2021-02-21
Detect host operating system in Makefile 

ChristianMoesl:
I had problems with Selfie on Windows 10, when compiled with Mingw64 (gcc). Selfie did not detect, that it runs on Windows, and therefore had problems to generate output files, because it used the wrong flags for `open()`.

So I figured, I should think of a more reliable way to detect, if Selfie is running on Windows. This is a more straightforward way to detect that, which could also be triggered on MacOs/Linux, if someone would use a ".exe" file extension on these platforms. But I think it is safe to say, that we probably never will see an issue with that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,25 @@
+# Give Selfie information about the operating system ($OS is defined in Windows only)
+ifeq ($(OS),Windows_NT)
+	HOSTOS := 3
+else
+	ifeq ($(shell uname -s),Darwin)
+		HOSTOS := 2
+	else ifeq ($(shell uname -s),Linux)
+		HOSTOS := 1
+	else
+		$(error operating system is not supported to build Selfie)
+	endif
+endif
+
 # Compiler flags
-CFLAGS := -Wall -Wextra -O3 -m64 -D'uint64_t=unsigned long long'
+CFLAGS := -Wall -Wextra -O3 -m64 -D'uint64_t=unsigned long long' -D'HOSTOS=$(HOSTOS)'
 
 # Bootstrap selfie.c into selfie executable
 selfie: selfie.c
 	$(CC) $(CFLAGS) $< -o $@
 
 # 32-bit compiler flags
-32-BIT-CFLAGS := -Wall -Wextra -Wno-builtin-declaration-mismatch -O3 -m32 -D'uint64_t=unsigned long'
+32-BIT-CFLAGS := -Wall -Wextra -Wno-builtin-declaration-mismatch -O3 -m32 -D'uint64_t=unsigned long' -D'HOSTOS=$(HOSTOS)'
 
 # Bootstrap selfie.c into 32-bit selfie executable, requires 32-bit compiler support
 selfie-32: selfie.c
--- a/selfie.c
+++ b/selfie.c
@@ -264,6 +264,8 @@ uint64_t WINDOWS_O_BINARY_CREAT_TRUNC_WRONLY = 33537;
 // these flags seem to be working for LINUX, MAC, and WINDOWS
 uint64_t S_IRUSR_IWUSR_IRGRP_IROTH = 420;
 
+uint64_t O_CREAT_TRUNC_WRONLY = 0; // write-only flags for HOSTOS initialised in init_system()
+
 // ------------------------ GLOBAL VARIABLES -----------------------
 
 uint64_t number_of_written_characters = 0;
@@ -412,6 +414,7 @@ uint64_t SYM_GEQ          = 27; // >=
 uint64_t SYM_INT      = 28; // int
 uint64_t SYM_CHAR     = 29; // char
 uint64_t SYM_UNSIGNED = 30; // unsigned
+uint64_t SYM_HOSTOS   = 31; // HOSTOS
 
 uint64_t* SYMBOLS; // strings representing symbols
 
@@ -447,7 +450,7 @@ uint64_t source_fd   = 0; // file descriptor of open source file
 // ------------------------- INITIALIZATION ------------------------
 
 void init_scanner () {
-  SYMBOLS = smalloc((SYM_UNSIGNED + 1) * SIZEOFUINT64STAR);
+  SYMBOLS = smalloc((SYM_HOSTOS + 1) * SIZEOFUINT64STAR);
 
   *(SYMBOLS + SYM_INTEGER)      = (uint64_t) "integer";
   *(SYMBOLS + SYM_CHARACTER)    = (uint64_t) "character";
@@ -481,6 +484,7 @@ void init_scanner () {
   *(SYMBOLS + SYM_INT)      = (uint64_t) "int";
   *(SYMBOLS + SYM_CHAR)     = (uint64_t) "char";
   *(SYMBOLS + SYM_UNSIGNED) = (uint64_t) "unsigned";
+  *(SYMBOLS + SYM_HOSTOS)   = (uint64_t) "HOSTOS";
 
   character = CHAR_EOF;
   symbol    = SYM_EOF;
@@ -1101,8 +1105,6 @@ void     implement_openat(uint64_t* context);
 void     emit_malloc();
 uint64_t try_brk(uint64_t* context, uint64_t new_program_break);
 void     implement_brk(uint64_t* context);
-
-uint64_t is_boot_level_zero();
 
 // ------------------------ GLOBAL CONSTANTS -----------------------
 
@@ -1750,6 +1752,8 @@ void print_register_memory_profile();
 
 void print_profile(uint64_t* context);
 
+void print_host_os();
+
 // ------------------------ GLOBAL CONSTANTS -----------------------
 
 uint64_t INSTRUCTIONSIZE       = 4;  // in bytes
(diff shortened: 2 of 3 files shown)


ckirsch (changes requested):
Nice idea! Can you check why it fails for the Windows target?

selfie.c
```diff
@@ -2308,13 +2310,35 @@ void init_system() {
...
-    // caution: the name of the executable must not have an extension to make this work
-    // try opening executable with zeroed flags which likely fails but just on Windows
-    if (signed_less_than(sign_extend(open(selfie_name, 0, 0), SYSCALL_BITWIDTH), 0))
+    // caution: this could also be true for executables on Linux/MacOs
+    if (has_exe_file_extension(selfie_name))
       WINDOWS = 1;
   }
 }
 
+uint64_t has_exe_file_extension(char* name) {
```
ckirsch, line 2319: This code repeats some work done in `replace_extension`. Can the two procedures be merged?

selfie.c
```diff
@@ -2308,13 +2310,35 @@ void init_system() {
   if (is_boot_level_zero()) {
     BOOTLEVELZERO = 1;
 
-    // caution: the name of the executable must not have an extension to make this work
-    // try opening executable with zeroed flags which likely fails but just on Windows
-    if (signed_less_than(sign_extend(open(selfie_name, 0, 0), SYSCALL_BITWIDTH), 0))
+    // caution: this could also be true for executables on Linux/MacOs
```
ckirsch, line 2313: We should probably include a tiny bit of information on the system selfie detected here to be sure it works:

./selfie: this is the selfie system from selfie.cs.uni-salzburg.at with
./selfie: 64-bit unsigned integers and 64-bit pointers on boot level 0

ChristianMoesl:
> Nice idea! Can you check why it fails for the Windows target?

I'll have a look at it on Saturday

ChristianMoesl:
@ckirsch It looks like executables on windows can be called with unix notation "./selfie" (happens in Makefile) and that is reflected in the first argument of the executable. It is not always the case, and therefore I did not notice it initially.

I pivoted. I detect the host operating system in the Makefile. It seems like a very clean fix, what do you think?

Makefile
```diff
@@ -1,12 +1,23 @@
 # Compiler flags
 CFLAGS := -Wall -Wextra -O3 -m64 -D'uint64_t=unsigned long long'
 
+# Give Selfie information about the host operating system
+ifeq ($(OS),Windows_NT)
+    CFLAGS += -D'HOST_OS=3ull'
+else
```
ckirsch, line 7: Can this be done in a single line simply passing $OS? $OS on my system is "OSX". Even if this doesn't work I would still prefer if we could do this just based on $OS and not additional stuff like `uname`.

Makefile
```diff
@@ -1,12 +1,23 @@
...
+    endif
+endif
+
 # Bootstrap selfie.c into selfie executable
 selfie: selfie.c
 	$(CC) $(CFLAGS) $< -o $@
 
 # 32-bit compiler flags
-32-BIT-CFLAGS := -Wall -Wextra -Wno-builtin-declaration-mismatch -O3 -m32 -D'uint64_t=unsigned long'
+32-BIT-CFLAGS := $(subst -D'uint64_t=unsigned long long', -D'uint64_t=unsigned long', $(subst -m64, -m32, $(CFLAGS))) -Wno-builtin-declaration-mismatch
```
ckirsch, line 20: How about defining HOSTOS and then use that in CFLAGS and 32-BIT-CFLAGS? This substitution business here is just an overkill.

selfie.c
```diff
@@ -5631,6 +5626,14 @@ void selfie_compile() {
     emit_fetch_data_segment_size_interface();
   }
 
+  data_size = data_size + SIZEOFUINT64;
+
+  // emit global HOST_OS variable for bootstrapping
+  create_symbol_table_entry(GLOBAL_TABLE, string_copy("HOST_OS"), 1, VARIABLE, UINT64_T, SELFIE, -data_size);
```
ckirsch, line 5632: This is a problem since it introduces another global variable that shifts all offsets of all other global variables. How about modifying `identifier_or_keyword` to rewrite `HOSTOS` to `SELFIE`? Without `_` it's the same number of characters...

selfie.c
```diff
@@ -6582,21 +6585,17 @@ uint64_t open_write_only(char* name) {
...
-  else {
-    // try Mac flags first as default
-    fd = sign_extend(open(name, MAC_O_CREAT_TRUNC_WRONLY, S_IRUSR_IWUSR_IRGRP_IROTH), SYSCALL_BITWIDTH);
+  uint64_t flags;
 
-    if (signed_less_than(fd, 0))
-      // then try Linux flags
-      fd = sign_extend(open(name, LINUX_O_CREAT_TRUNC_WRONLY, S_IRUSR_IWUSR_IRGRP_IROTH), SYSCALL_BITWIDTH);
-  }
+  if (HOST_OS == WINDOWS)
```
ckirsch, line 6590: We should set the correct flags in `init_system` now and then just use those.

selfie.c
```diff
@@ -7148,11 +7147,15 @@ void implement_openat(uint64_t* context) {
   mode      = *(get_regs(context) + REG_A3);
 
   if (down_load_string(context, vfilename, filename_buffer)) {
-    if (flags == MAC_O_CREAT_TRUNC_WRONLY)
-      // default for opening write-only files
-      fd = open_write_only(filename_buffer);
-    else
-      fd = sign_extend(open(filename_buffer, flags, mode), SYSCALL_BITWIDTH);
+    if (flags == LINUX_O_CREAT_TRUNC_WRONLY) {
```
ckirsch, line 7150: Same as above

ckirsch (changes requested):

Makefile
ChristianMoesl: Good point, I'll try that.

Makefile
ChristianMoesl: I liked that, because it emphasises, that 32 bit flags are derived from 64 bit flags. But anyway I have no hard opinion on that, so I'll revert that change.

selfie.c
ChristianMoesl: I do not understand, why that would be a problem. We do this also with `_bump`. Or am I missing something?

selfie.c
ChristianMoesl: Yeah sure

ChristianMoesl:
@ckirsch I changed it, such that we pass a String as HOST_OS flag. Unfortunately to get the OS string, we need `uname`. There is no $OS environment variable in every Linux distribution, but there is `uname` in all Unix OS (https://en.wikipedia.org/wiki/List_of_Unix_commands). Also strangely on MacOS, it doesn't give me a value for `$(OS)` in the Makefile, even tough there is an environment variable $OS="OSX" set in the console.

One problem I had with that implementation is to bootstrap the HOST_OS variable. Because it has to have an address of an string literal in the symbol table. That also means, this address is dependent on the global pointer. Therefore this address has to be computed in bootstrapping.

Makefile
```diff
@@ -1,12 +1,18 @@
+# Give Selfie information about the operating system ($OS is defined on Windows only)
+ifndef OS
+	# use uname or fail with "Unknown"
+	OS := $(shell sh -c 'uname 2>/dev/null || echo Unknown')
```
ckirsch, line 4: This is nice but in the end I think the code you used before was better. How about going back to the code before but fail if the OS cannot be determined to be one of the three supported OSs? Pass to selfie just integers, as before.

Makefile
```diff
@@ -1,12 +1,18 @@
...
+endif
+
 # Compiler flags
-CFLAGS := -Wall -Wextra -O3 -m64 -D'uint64_t=unsigned long long'
+CFLAGS := -Wall -Wextra -O3 -m64 -D'uint64_t=unsigned long long' -D'HOST_OS="$(OS)"'
```
ckirsch, line 8: You can keep that but pass in an integer that identifies the OS. This avoids any string dependencies in selfie.

selfie.c
```diff
@@ -5463,6 +5470,11 @@ void emit_bootstrapping() {
   // code segment starts at PK_CODE_START
   code_start = PK_CODE_START;
 
+  // store HOST_OS string literal for Selfie operating system
```
ckirsch, line 5473: This is too complicated. What I meant is to bootstrap in the selfie compiler all occurrences of `HOST_OS` or better `HOSTOS` to `SELFIE`. And `SELFIE` should be a global "constant" indicating that the system is hosted by selfie.

ckirsch: Code like `HOSTOS == SELFIE` then becomes `SELFIE == SELFIE`.

ckirsch (changes requested):

Makefile
ChristianMoesl: Done

selfie.c
ChristianMoesl: Got it

Makefile
ChristianMoesl: done

ChristianMoesl:
@ckirsch I changed it back to the version with integers. I also am replacing `HOSTOS` with an integer literal `0` (= SELFIE) in in `identifier_or_keyword`. Therefore we are consistent with these preprocessor macros, who are also replacing HOSTOS with a integer literal. And we do not run into problems in sources, where the `SELFIE` identifier is not available (e.g. examples).

I think this version is the best out of all of them. Simple and clean.

PR #262, opened 2021-03-16
Run Docker Selfie github action modified so it runs only on original …

nfejzic:
…repository.


ckirsch:
@nfejzic Thanks!

nfejzic:
@ckirsch You're welcome!

PR #264, opened 2021-03-17
Fix qemu and spike target for printf

Son-Typ:
Our stack bootstrapping was wrong due to a misunderstanding on our part. This PR correctly bootstraps the stack so that qemu and spike are able to run.


ckirsch:
Thanks a lot!

PR #269, opened 2021-04-01
Use gcc from cygwin to build on windows

mstarzinger:
This installs `gcc` and `make` from Cygwin to build on Windows. I did not add a step to run the autograder baseline, but we could try whether using `python` from Cygwin works at some point.


ckirsch:
@mstarzinger Thanks a lot!

@ChristianMoesl What do you think? Can you get `make grade` to work on Windows using @mstarzinger's code?

PR #270, opened 2021-04-02
Machine: Implement `dprintf` and `sprintf`

Blazefrost:
This commit introduces functions `dprintf` and `sprintf` to the kernel's tiny C library. This is required because the `printf` branch introduces dependencies on aforementioned functions as well as `printf`, which has partially been implemented for use by the kernels already.
As the kernels build in a freestanding environment, no C standard library is provided to us and we must provide relevant functions ourselves.

This commit does only fix compilation failures, but does not yet implement relevant format string flags.

### DIFF
--- a/machine/filesystem.c
+++ b/machine/filesystem.c
@@ -16,3 +16,35 @@ const KFILE* find_file(const char* filename) {
   else
     return NULL;
 }
+
+bool fd_is_stdio(int fd) {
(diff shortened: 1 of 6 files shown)


ckirsch:
@Blazefrost Thanks so much! Does this PR already work with the `make --directory machine/` command in the docker file?

Issue #271, opened 2021-04-03
`characters of assembly` count in disassembler output is incorrect in `printf` branch


ckirsch:

ckirsch:
@Son-Typ Could you please have a look and see if you can fix this? Thanks!

Son-Typ:
A small test case would be useful. Is the `number_of_written_characters` wrong? I tried compiling using `./selfie -c selfie.c -o selfie.m` and `./selfie -l selfie.m -m 42 -c selfie.c -o selfie2.m` and those are my results:

For `main`:

- bootlevel 0: 185520 bytes generated with 42906 instructions and 13896 bytes of data
- bootlevel 1: 282792 bytes generated with 42906 instructions and 13896 bytes of data

For `printf`:

- bootlevel 0: 184936 bytes generated with 42580 instructions and 14616 bytes of data
- bootlevel 1: 184936 bytes generated with 42580 instructions and 14616 bytes of data

The `printf` branch generates the same number of bytes for both bootlevels which I would expect to make more sense.

What are the expected results?

ckirsch:
@Son-Typ I can't reproduce the issue you are seeing for `main`. I am on commit 0de776c6.

What I actually meant is the character count, not the byte count. But we may have to check both.

Try: `./selfie -c selfie.c -S selfie.s`

Should show more than a million characters in the assembly file but does not on the printf branch. I am on commit 0cbbb300.

Son-Typ:
I was on the same commits while conducting both tests.
Regarding the problem: The reason is actually quite simple. At bootlevel 0 we use the bootstrapping compiler's `dprintf` to write into the file, while we use our own `dprintf` at the other bootlevels. Clang's/GCC's `dprintf` does not count the number of written characters in `number_of_written_characters` while the `printf` implementation in `main` and our `dprintf` implementation do count. We probably want to count the written characters inside `direct_output` as `dprintf` returns the number of written characters. In this case we would have to stop counting inside `put_character`, which is a nasty side effect to begin with.
I tried `./selfie -l selfie.m -c selfie.c -S selfie.s`, i.e., disassembling at bootlevel 1, which resulted in the expected "selfie.m: 1901357 characters of assembly [...]".

MartinGroesbacher:
We will work on it asap but since its family holidays it might take a little while. Will post here when we're done

ckirsch:
Sure, that's fine. Thanks a lot for looking into this. I hope there is a clean solution.

[...]

PR #273, opened 2021-04-08
Fix char

MartinGroesbacher:
Suggestive solution regarding #271


ckirsch:
@MartinGroesbacher Thanks!

PR #275, opened 2021-04-21
Update grader to show executed commands at all times

CAS-ual-TY:
As suggested in the lecture, someone (me) made the grader spit out what it is executing.
It was already doing that for failed tests, now it does it all the time. Also colored in yellow/orange/gold.

picture

(ignore the lines above the first [PASSED], its my own script I use so I do not have to remember exact assignment identifiers)

### DIFF
--- a/grader/lib/checks.py
+++ b/grader/lib/checks.py
@@ -9,7 +9,7 @@ from .model import Check, CheckResult
 from .output_processing import (filter_status_messages, has_compiled,
                                 has_no_compile_warnings, has_no_bootstrapping_compile_warnings, 
                                 is_interleaved_output, is_permutation_of)
-from .print import print_processing, print_warning, stop_processing_spinner
+from .print import print_processing, print_warning, stop_processing_spinner, print_command
 from .system import INSTRUCTIONSIZE, WORDSIZE, read_data, read_instruction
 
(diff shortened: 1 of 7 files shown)


CAS-ual-TY:
Note: print_passed is also called in grade.py on line 16. But this seems to never be called (it also prints "blaaaaa" so I assume it is unused or forgotten about or maybe still indev) so I ignored it.

ckirsch:
@CAS-ual-TY Thanks a lot! This is very nice. How about changing the order: print the command first and then whether it passed or not? Can you do that?

@ChristianMoesl What do you think about this? Also, please check out the "blaaaa" print :) I guess we don't need that anymore.

CAS-ual-TY:
@ckirsch Done and committed. I had to slightly expand the current model to gain access to the executed command (before it is actually executed).

picture

CAS-ual-TY:
Another alternative formatting suggestion:

picture

ckirsch:
@CAS-ual-TY Even better.

ChristianMoesl (approved):
Looks good to me! 🚀

ckirsch:
@ChristianMoesl What about the `blaaa` in the grader somewhere? Could you please fix that?

PR #276, opened 2021-04-30
Handling pointer dereference of an identifier as a special case of dereferencing an expression.

nfejzic:
At the lecture on 28th April we've checked out how are pointers handled in selfie.

As mentioned in the lecture by @ckirsch, `*x` could potentially be handled as a special case of dereferencing an expression `*(x + 0)`.

This is my proposal on how this case could be handled.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -4981,6 +4981,7 @@ void compile_statement() {
   char* variable_or_procedure_name;
   uint64_t* entry;
   uint64_t offset;
+  uint64_t pointer_is_expression;
 
   // assert: allocated_temporaries == 0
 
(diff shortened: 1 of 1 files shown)


selfie.c
```diff
@@ -4997,80 +4998,58 @@ void compile_statement() {
...
     get_symbol();
 
-    // "*" variable
-    if (symbol == SYM_IDENTIFIER) {
-      ltype = load_variable_or_big_int(identifier, VARIABLE);
-
-      if (ltype != UINT64STAR_T)
-        type_warning(UINT64STAR_T, ltype);
-
+    // "*" "(" expression ")"
```
ckirsch, line 5001: I am wondering if you can simply parse `* expression` without the surrounding parentheses. Also, please check what the official C grammar is in this case.

nfejzic: I am not sure what do you exactly mean. Do you mean to parse for example `*x+1` as `*(x+1)`?

ckirsch: The question is whether the parentheses are required. Can you check the official C grammar for that?

nfejzic: I just checked it in [the ISO standard for C, page 843](http://www.open-std.org/JTC1/SC22/WG14/www/docs/n1570.pdf#page=483), and as far as I understand the specification the parentheses are needed when dereferencing an expression.

It's defined in `unary-expression`, where among others, one possibility is `unary-operator cast-expression`. This is further specified recursively, but in the end when `expression` appears it is surrounded by parentheses. (See `primary-expression`). 

i.e. type casting does not need parentheses: `*(char)x`, but expression does `*(x+1)`.

ckirsch: Got it! Thank you. I am afraid that means there is not much we can do here. Your code is fine but doesn’t really make things simpler. Let’s keep it as is.

PR #281, opened 2021-05-06
Fix assignment listing order, closes #278

### LINKED ISSUE #278: Order in list of assignments out of sync
In https://github.com/cksystemsteaching/selfie/blob/main/assignments/compiler-assignments.md the assignment `for-loop` is the very last one. Yet in https://github.com/cksystemsteaching/selfie/blob/main/grader/self.py#L323 that assignment (when listing all assignments) is shown between `bitwise-and-or-not` and `array`.

Possibly in the past the assignment was reordered and the order in the other list was not updated.

It would be good to have the same order, such that as a student there is less confusion. This is important when working on future assignments, such that you can do them in the right order having fewer rebasing conflicts to worry about later.

kim366:


ChristianMoesl (approved):
looks good

PR #283, opened 2021-05-08
Grader: Fix ELF program header code start/length parsing (closes #277)

### LINKED ISSUE #277: Grader not terminating on plain selfie
When executing the grader with `./grader/self.py bitwise-shift-execution` on plain selfie without any modifications the test `right-shift has right RISC-V encoding` does not terminate. Expected behaviour would be grade 5.

Blazefrost:
Somewhere around commit 0bd8da6, Selfie changed the way the ELF header
is emitted. Instead of a combined section, there are two separate
sections, one for code and one for data. By changing the header, the
grader did read the code start/length from the wrong location.

As before, the field parsing is hardwired to Selfie's output to keep
complexity minimal (otherwise we would require a full-fledged ELF
parser)

There's one caveat, though: Selfie now supports 32-bit as well. However,
the grader is not able to parse ELF32 binaries as of now because the
length of Elf64_Off/Elf64_Addr and Elf32_Off/Elf32_Addr differ.

### DIFF
--- a/grader/lib/checks.py
+++ b/grader/lib/checks.py
@@ -126,15 +126,26 @@ def check_instruction_encoding(instruction, file) -> List[Check]:
 
                 try:
                     with open('.tmp.bin', 'rb') as f:
-                        ignored_elf_header_size = 14 * WORDSIZE
+                        # Read the code start/length out of the program header
+                        # (p_filesz in the ELF program header)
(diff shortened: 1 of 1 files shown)


ckirsch:
@Blazefrost Thanks a lot! 64-bit support only is perfectly fine in this case.

PR #284, opened 2021-05-10
Check for compiler warnings during assignment compilation

ChristianMoesl:
Because Selfie will not return with an status_code != 0 for
compiler warnings, the grader should also check for warnings
in the output during assignment compilation.


Blazefrost (approved):
Oh, I wasn't aware that the grader already had a method to check for compilation warnings!

Looks fine :+1:

PR #288, opened 2021-06-17
Machine: Move RISC-U binary artifacts to build directory (closes #287)

### LINKED ISSUE #287: Residual files in selfie root directory in the selfie docker image after machine build
@Blazefrost could you please help me in fixing the following issue:

The selfie root directory in the selfie docker image contains two residual files called `hello-world.m` and `selfie.m` which, I believe, are created during `RUN make --directory machine/`. I think it would be better if those two files are removed somehow or put in the `machine` folder.

What do you think?

Blazefrost:
Instead of populating the repository root while performing a machine build, place the RISC-U binaries that are packaged into machine's build directory.

### DIFF
--- a/machine/Makefile
+++ b/machine/Makefile
@@ -31,6 +31,12 @@ include defines.mk
 include filesystem.mk
 
 # ==============================================================================
+# Add rules for packaged RISC-U binaries
+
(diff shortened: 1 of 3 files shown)


ckirsch:
@Blazefrost Thanks a lot!

PR #289, opened 2021-06-28
Implement break & continue

CAS-ual-TY:
Hi,
as written in a private message, I have implemented break and continue in my private repo and now offer to port that over to the main selfie repo.
It is implemented similar to how the return works using fixup chains. Additionally, the compiler knows if it is currently inside of a loop or not. Whenever a loop is compiled that is inside of another loop, the fixup chains are temporarily stored and then later restored at the end of the inner loop.

I can also adapt the for-loop assignment to check for successful implementation for break and continue if you would be interested in that. In my opinion, the assignment does not really get harder since including that in a for-loop implementation is very easy todo and barely different from how it is done by the while-loop implementation.

I have also updated the gammar. I incorporated the fact that break and continue can only be used inside of a while loop, which is why I did not add them to the actual statement definition, but I could of course also do that if its better that way.

If there is anything else needed or any issues, please let me know.

Tested compilation using both GCC and CLANG on a linux debian subsystem running on windows 10.

### DIFF
--- a/examples/break.c
+++ b/examples/break.c
@@ -1,0 +1,37 @@
+uint64_t x;
+
+uint64_t main() {
+  x = 32;
+
+  while (1) {
+    x = x + 2;
+
+    break;
+
+    x = x + 100;
+  }
+
+  while (1) {
+    while (1) {
+      x = x + 3;
+
+      break;
+
+      x = x + 100;
+    }
+
+    x = x + 5;
+
+    break;
+
+    x = x + 100;
+  }
+
+  while (1) {
+    break;
+
+    x = x + 100;
+  }
+
+  return x; // 42
+}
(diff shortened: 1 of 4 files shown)


ckirsch:
@CAS-ual-TY Thanks a lot for your PR! I should have replied earlier, sorry about that. Your code looks nice and you are welcome to PR it against a branch I just created called `break`. However, I may not be able to pull it into the `main` branch because `break` and `continue` make C* more complex and thus make teaching C* more difficult without really being needed in the implementation of selfie. For example, selfie still does not actually use these statements. That contradicts the spirit of the project which is self-referentiality. There are, however, interesting, more advanced projects such as support of `char` that I would be happy to advise you on.

CAS-ual-TY:
Makes sense! I switched the branch, everything should be good

PR #292, opened 2021-06-30
Fix execution depth tracking in monster.

mstarzinger:
This ensures the current execution depth of the context is updated
after each context switch. The maximum execution depth specified at the
command line is now properly respected when checking for timeouts.


ckirsch:
@mstarzinger Thanks!

PR #299, opened 2021-10-15
Simplify merging of symbolic memory stores.

mstarzinger:
This unifies the previous two slightly distinct merging loops into a
single loop in `merge_symbolic_memory`. Note that we now iterate over
the entire store of the active context. The previous assumption that
diverging values can only appear in the unshared portion does not hold,
when non-local control-flow (i.e. returns nested in loops) is present
within the function body. This is both simpler and fixes several bugs.

Additional minor changes aside from the above algorithmic change:
- The special `MERGED` marker address is no longer needed.
- Missing updates of `additional_memory` in the loop were fixed.
- Call `update_begin_of_shared_symbolic_memory` earlier.

Note that overall complexity of the merging is still in O(n^2) where n
is the maximum size of the symbolic memory store, same as before.


ckirsch:
@mstarzinger Awesome! Thanks a lot!

PR #312, opened 2022-03-14
New grading tasks: logical-and-or-not and lazy-evaluation

CAS-ual-TY:
**logical-and-or-not**
A bunch of different boolean expressions are tested to cover as many error-cases as I was able to find/think of.

**lazy-evaluation**
Built on logical operators of course. The grading checks if basic boolean expressions still work followed by if lazy evaluation works by having endless loops in conditions which should be skipped if lazy evaluation was implemented successfully.

### DIFF
--- a/grader/assignments/lazy-eval/advanced-logical-expressions.c
+++ b/grader/assignments/lazy-eval/advanced-logical-expressions.c
@@ -1,0 +1,29 @@
+int main(int argc, char** argv) {
+  uint64_t a;
+  uint64_t b;
+  uint64_t c;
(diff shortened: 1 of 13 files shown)


ckirsch:
@CAS-ual-TY Awesome! Thanks a lot. Any clue why the docker build fails?

CAS-ual-TY:
Yes. I made a trivial mistake which I believe is the cause of it. Lets see if it works now.

CAS-ual-TY:
Yup, I was right. This is good to go!

PR #313, opened 2022-03-22
Workflow improvements

CAS-ual-TY:
Workflow improvements as discussed on monday.

**The Problem for students**
Specific parts of the current workflows eat up a lot of quota to a point where you can run all workflows maybe 3 times on total before you run out of avaiable minutes. This seems to be a problem of this semester's lecture (according to Michael) and has already been a problem of the same lecture when I took part in it last year.

**Solutions**
Make the running of jobs dependent on the OS the are run on (eg. 1 min on Linux = 10 min on MacOS), the repo they are in and if their repo is public or private, and the branch they are run on.

**Extras**
Add manual dispatchers to all workflows.

**Task List**
- [x] Disable jobs that run on Windows or MacOS since these are more "expensive" in minutes for private repos only
- [x] Disable jobs for all branches except the main branch for private repos only (still TODO. The idea is that only the final squash merge should run anything online, while any still-in-dev commites on dev branches should not eat up any quota)
- [x] Add manual dispatcher for the "Make everything selfie" workflow
- [x] Add manual dispatcher for the grader workflow and add an input field to determine the grader task (eg. hex-literal)
- [x] Disable scheduled run for private repos only

### DIFF
--- a/.github/workflows/grader.yml
+++ b/.github/workflows/grader.yml
@@ -19,6 +19,7 @@ jobs:
     env:
       COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
       ASSIGNMENT: ${{ github.event.inputs.assignmentid }}
+    if: ${{ (!github.event.repository.private) || (github.ref == 'refs/heads/main') }}
 
     steps:
     - name: Checkout selfie
--- a/.github/workflows/grader.yml
+++ b/.github/workflows/grader.yml
@@ -8,9 +8,32 @@ on:
   workflow_dispatch:
     inputs:
       assignmentid:
-        description: 'Assignment ID'     
+        description: 'Assignment ID'
+        type: choice
         required: true
-        default: 'self-compile'
+        options:
+        - bootstrapping
+        - self-compile
+        - print-your-name
+        - hex-literal
+        - bitwise-shift-compilation
+        - bitwise-shift-execution
+        - bitwise-and-or-not
+        - array
+        - array-multidimensional
+        - struct-declaration
+        - struct-execution
+        - for-loop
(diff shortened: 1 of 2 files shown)


CAS-ual-TY:
**Note**
The manual dispatcher for the "Make everything selfie" target will also ignore Windows and MacOS on private branches. But I could probably add an checkbox input for manual dispatch, where you can choose which OS to run it on.

CAS-ual-TY:
Just the last task is missing. I can also make the input field for a manually dispatched grader run a dropdown menu, instead of a text input (see https://stackoverflow.com/questions/69296314/dropdown-for-github-workflows-input-parameters).
The disadvantage would be that this list would have to be maintained in case new assignments get added.

CAS-ual-TY:
I would have this specific change ready in case you want me to change it (dropdown of grading targets):

grafik

ckirsch:
> **Note**
> The manual dispatcher for the "Make everything selfie" target will also ignore Windows and MacOS on private branches. But I could probably add an checkbox input for manual dispatch, where you can choose which OS to run it on.

That would be nice!

ckirsch:
> I would have this specific change ready in case you want me to change it (dropdown of grading targets):
> 
> grafik

sure, add it. thanks!

CAS-ual-TY:
- OS/Jobs are now selectable for manual dispatches
- Private forks still ignore windows/macOS runs AND ignore any runs on non-main branches EXCEPT when there is a manual dispatch
- This is the case for both the grader workflow and the make everything selfie workflow
- Tested on my own public fork and a private fork (I sent you an invite in case you want to test it)

Merge is ready I believe.

PR #317, opened 2022-04-25
Add context struct size to lib - make monster and boehm-gc context extension dynamic

CAS-ual-TY:
As discussed in the meeting.
monster and boehm-gc should now dynamically adjust to any context struct size changes. I used 'CONTEXTENTRIES' as variable name because any 'SIZEOF' variable names would always be in bytes, and I did not want to break this convention.


CAS-ual-TY:
I tested the 'boehm-gc' and 'smt' make targets locally but did not 'make all' yet. Lets see if the workflow fails or succeeds.

PR #318, opened 2022-05-11
Showcase the use of expect_symbol as mentioned in the lecture.

user-0xcafe:
As I mentioned in the lecture today, the use of `expect_symbol`,
instead of the 4 lines that it replaces, can more clearly communicate 
what is happening in the code.

I am curious what you think about it @ckirsch.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -417,6 +417,7 @@ uint64_t identifier_string_match(uint64_t string_index);
 uint64_t identifier_or_keyword();
 
 void get_symbol();
+void expect_symbol(uint64_t expected_symbol);
 
 void handle_escape_sequence();
 
@@ -3931,6 +3932,13 @@ void get_symbol() {
   }
 }
 
+void expect_symbol(uint64_t expected_symbol) {
+  if (symbol == expected_symbol)
+    get_symbol();
+  else
+    syntax_error_symbol(expected_symbol);
+}
+
 void handle_escape_sequence() {
   // ignoring the backslash
   number_of_ignored_characters = number_of_ignored_characters + 1;
@@ -4728,19 +4736,13 @@ uint64_t compile_factor() {
 
       cast = compile_type();
 
-      if (symbol == SYM_RPARENTHESIS)
-        get_symbol();
-      else
-        syntax_error_symbol(SYM_RPARENTHESIS);
+      expect_symbol(SYM_RPARENTHESIS);
 
     // not a cast: "(" expression ")"
     } else {
       type = compile_expression();
 
-      if (symbol == SYM_RPARENTHESIS)
-        get_symbol();
-      else
-        syntax_error_symbol(SYM_RPARENTHESIS);
+      expect_symbol(SYM_RPARENTHESIS);
 
       // assert: allocated_temporaries == n + 1
 
@@ -4825,10 +4827,7 @@ uint64_t compile_factor() {
 
     type = compile_expression();
 
-    if (symbol == SYM_RPARENTHESIS)
-      get_symbol();
-    else
-      syntax_error_symbol(SYM_RPARENTHESIS);
+    expect_symbol(SYM_RPARENTHESIS);
   } else {
     syntax_error_unexpected();
 
--- a/selfie.c
+++ b/selfie.c
@@ -417,8 +417,8 @@ uint64_t identifier_string_match(uint64_t string_index);
 uint64_t identifier_or_keyword();
 
 void get_symbol();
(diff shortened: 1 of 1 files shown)


ckirsch:
@user-0xcafe Very nice! Please go ahead and apply that everywhere. I will then pull that in after the current assignment.

nbes4:
Little bit of nitpicking here, but since this addition is meant for more code clarity, would it make sense to rename `expect_symbol` to something like `expect_symbol_and_get()`, `expect_and_get_symbol()` or `expect_symbol_and_consume()` etc? Since, especially for "newcomers", `expect_symbol()` sounds more like a check, but it eventually consumes the current symbol and moves on to the next.

Any opinions on this?
Otherwise, great addition 👍

user-0xcafe:
Great question. I do like the some of your suggestions, especially those that read almost like sentences. 

One reason I didnt use a longer name as that I wanted to mimic `get_symbol`.

The other reason is that all of those names only tell half the story. They do not talk about emitting an error if the symbols do not match. I thought thet this mignt be more confusing. 

But I am happy to change it to `expect_and_get_symbol` if you think that I am overthinking this.

nbes4:
I understand your concerns and the goal to mimic `get_symbol`. In the meantime the name `expect_symbol()` kind of implies that it can fail if we get something that we not "expect", so adding `and_get` would surely be a small bonus for clarity. However, this is just my subjective thinking, so please make sure you name the function as you see fit.

ckirsch:
> I believe that with the second commit I converted all the “low-hanging fruits”. There is a slight variation of the theme that I didn't touch: when the parser exits on a syntax error.

How about another procedure for this case? How often does it occur? How about `force_symbol()` for that?

user-0xcafe:
@ckirsch I just implemented that in the third commit. 

@nbes4 Another possible name that I considered are `get_expected_symbol` and `get_required_symbol`. I would like to get your opinion on those.

ckirsch:
> @ckirsch I just implemented that in the third commit.
> 
> @nbes4 Another possible name that I considered are `get_expected_symbol` and `get_required_symbol`. I would like to get your opinion on those.

@user-0xcafe Nice! Please use those names instead.

nbes4:
`get_expected_symbol `and `get_required_symbol` seem like a good and clear choice!

user-0xcafe:
@ckirsch I changed the names and am very pleased with the way it reads. 
@nbes4 Thanks for your input!

selfie.c
```diff
@@ -417,6 +417,8 @@ uint64_t identifier_string_match(uint64_t string_index);
 uint64_t identifier_or_keyword();
 
 void get_symbol();
+void get_expected_symbol(uint64_t expected_symbol);
+void get_required_symbol(uint64_t expected_symbol);
```
ckirsch, line 421: `expected_symbol` -> `required_symbol`

selfie.c
```diff
@@ -3931,6 +3933,23 @@ void get_symbol() {
...
 }
 
+void get_expected_symbol(uint64_t expected_symbol) {
+  if (symbol == expected_symbol)
+    get_symbol();
+  else
+    syntax_error_symbol(expected_symbol);
+}
+
+void get_required_symbol(uint64_t expected_symbol) {
```
ckirsch, line 3943: call `get_expected_symbol()` here but with a return value that indicates an error on which you then exit.

selfie.c
```diff
@@ -11846,4 +11805,4 @@ int main(int argc, char** argv) {
   exit_code = selfie(0);
 
   return exit_selfie(exit_code, " [ ( -m | -d | -r | -y ) 0-4096 ... ]");
-}
\ No newline at end of file
+}
```
ckirsch, line 11808: no trailing empty line please

ckirsch (changes requested):
@user-0xcafe great work! please fix the remaining issues, I will then pull.

selfie.c
```diff
@@ -3933,21 +3933,23 @@ void get_symbol() {
...
 
-void get_expected_symbol(uint64_t expected_symbol) {
-  if (symbol == expected_symbol)
+uint64_t get_expected_symbol(uint64_t expected_symbol) {
+  uint64_t found;
+
+  if (symbol == expected_symbol) {
     get_symbol();
-  else
+    found = 1;
```
ckirsch, line 3941: Just `return 1` here, no need for `found`.

ckirsch (changes requested):

PR #319, opened 2022-05-16
Thread-safe malloc grading task

CAS-ual-TY:
Quickly added the grading for this task.

This is ready for merge.

### Previously

`treiber-stack` task:
- check lr/sc RISC-V encoding
- check lr/sc assembly instruction format
- test push() and check if everything is in memory
- test pop() and check if everything pushed is popped correctly

### Now

`threadsafe-malloc` task:
- check lr/sc RISC-V encoding
- check lr/sc assembly instruction format
- test if syscall brk (on malloc) does not force a context switch anymore
- test thread-safe malloc

`treiber-stack` task (requires passing `threadsafe-malloc`):
- test push() and check if everything is in memory
- test pop() and check if everything pushed is popped correctly

## How it works

Right now malloc() does the following:

- load (get bump pointer)
- ecall (syscall brk: increase program break)
- store (set bump pointer to new program break)

However, syscalls currently force a context switch. This means that if thread 1 does malloc followed by thread 2 doing a malloc, they both get the same address resulting in an overwrite one way or another. This is what is being tested and is also required to be solved (very simple to do).

## Notes

I said earlier today that this breaks older assignments. Well, this is only the case if you make syscalls *not* force a context switch in general, instead of just doing that for brk. By trying a new implementation all older assignments still work without having to change anything that is related to any syscall other than brk (as in: it is not much work).
TL;DR: Older assignments do not break.

And on another note: Since the old treiber-stack grading task was "only" testing on Mipster (not Hypster), I decided to keep it that way. Another reason being that the threadsafe test requires a timing in instructions that I so far only achieved in Mipster.


CAS-ual-TY:
I need to change the `threadsafe-malloc.c` test to somehow force a context switch between the lr and sc calls. Probably by counting or guessing the amount of instructions left to execute and adding the malloc call to the end of that so lr is included but sc is not (timer related context switch before sc).
Maybe I can do something like compile code and output as binary, then change said binary to add a bunch of nops before the malloc-jal so the above happens and then load this "hacked" binary back in and execute it.

CAS-ual-TY:
Moved into the new `threadsafe-malloc-grading` branch

ckirsch:
@CAS-ual-TY there are some permission issues with self.py please fix.

ckirsch:
@CAS-ual-TY Looks good! Just keep the file permissions of Makefile as is. Should not be executable...

PR #324, opened 2022-06-21
Add hex-literal test to check if selfie can still parse 0-prefixed integer literals

fischer-martin:
I've seen that quite a few students seem to forget/ignore that selfie is able to parse 0-prefixed integer literals such as `05` when they are working on the `hex-literal` assignment. For some of them, this led to extremely weird bugs down the road where `0` literals weren't parsed in some circumstances.
Therefore, I've added a corresponding test (mandatory) to prevent situations like these. The literals in the test-file have been chosen in a ways such that a normal C-compiler such as `gcc` returns the same exit code as selfie (since in C 0-prefixed literals are normally interpreted in base 8 whereas in C* they are interpreted in base 10).


fischer-martin:
@ckirsch What do you mean exactly? I couldn't spot any changes in `get_symbol()` or `atoi()` that would remove support for 0-prefixed integer literals and would therefore prevent an expression like `05  == 5` or even `010 == 10` (which would evaluate to `0` when compiled with e.g. `gcc` due to the octal-prefix) from evaluating to `1` in selfie.

ckirsch:
@fischer-martin see 6d10353ede9e21147226db9170c3474d05c959e8 in a dev branch.

fischer-martin:
@ckirsch I see; I was only looking at the `main` branch. I guess in this case there isn't a need for this new test and we can close this PR.

ckirsch:
@fischer-martin sorry about that. Thank you anyway for pushing this.

PR #339, opened 2022-10-10
fix: Fixes 'may be uninitialized' warning reported by GCC version 12 with -O3 enabled.

danielkocher:
Remove `has_cast` variable and use properly initialized `cast` variable instead to distinguish whether a cast is encountered or not.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -5028,7 +5028,6 @@ uint64_t compile_term() {
 }
 
 uint64_t compile_factor() {
-  uint64_t has_cast;
   uint64_t cast;
   uint64_t type;
   uint64_t negative;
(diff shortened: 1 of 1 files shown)


ckirsch:
@danielkocher Thanks a lot. Had to do an alternative, cleaner fix though.

danielkocher:
@ckirsch I agree that having the `has_cast` variable benefits the readability. Unfortunately, the current fix does not resolve the problem.

Output of `gcc-12 -Wall -Wextra -O3 -m64 -D'uint64_t=unsigned long' -c selfie.c`:
```
selfie.c: In function ‘compile_factor’:
selfie.c:5032:12: warning: ‘cast’ may be used uninitialized [-Wmaybe-uninitialized]
 5032 |   uint64_t cast;
 ```
 
 I can issue another PR in the evening that preserves the `has_cast` variable and simply initializes `cast = 0;` in the first top-level `else` branch in `compile_factor`.

danielkocher:
This fix solves the problem. If I have some spare time, I try to figure out what `gcc-12` does (since this warning is only reported with `-O3` but not with `-O2` and lower optimization levels).

PR #369, opened 2024-01-29
feat: support bootloading of binaries produced by gcc

nfejzic:
# Work in progres...

This PR aims to extend the boot loader of `selfie` such that it can load binaries produced both by `selfie` and `gcc`. 

The changes introduced in this PR include reading the number of page header entries and reading them in separate page table. It seems like `gcc` produces three segments, one of which is machine specific `RISCV_ATTRIBUT`, and the other two are code and data segments. 

Boot loader now extracts needed information from ELF header, like entry, code segment offset in file and so on. This is necessary, because `gcc` sets these differently from selfie. 

## How to produce RISC-V Binary with `gcc`? 

To produce RISC-V Binary using `gcc` you need to install [RISC-V GNU Compiler Toolchain](https://github.com/riscv-collab/riscv-gnu-toolchain). You can do so with [Homebrew](https://brew.sh/) on macos, using this formula: [homebrew-riscv](https://github.com/riscv-software-src/homebrew-riscv).

Then, compile your binary using `riscv64-unknown-elf-gcc` binary. For example, you can compile `selfie` like this:

```bash
riscv64-unknown-elf-gcc -Wall -Wextra -D'uint64_t=unsigned long' -static selfie.c -o selfie_riscv
```

Then simply load the binary: 

```sh
# assuming you compiled selfie with 'make selfie'
./selfie -l selfie_riscv
```

If you try to run it in the current state, unknown instruction exception will be reported. For debugging purposes binary representation of the instruction is printed as well. You can copy the instruction and paste it in [RISC-V Instruction Encoder/Decoder](https://luplab.gitlab.io/rvcodecjs/) to see what the instruction is.

### DIFF
--- a/selfie.c
+++ b/selfie.c
@@ -1090,7 +1090,8 @@ void finalize_data_segment();
 
 uint64_t* touch(uint64_t* memory, uint64_t bytes);
 
-uint64_t* allocate_elf_header();
+uint64_t* allocate_elf_header(uint64_t with_page_header_table);
+uint64_t* allocate_elf_ph_table(uint64_t e_phentsize, uint64_t c);
 
 uint64_t* encode_elf_header();
 void      decode_elf_header(uint64_t* header);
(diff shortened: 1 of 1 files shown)


nfejzic:
@ckirsch this is now implemented in b8c9a77 (with follow-up commits). Should we close this PR?

PR #378, opened 2024-05-01
Add assembly example

thomaswulz:
The resulting binary works fine with both mipster and rotor.


ckirsch:
@thomaswulz Thanks a lot!

PR #380, opened 2024-05-15
Implement pseudoinstructions in Rotor disassembly

Jean-M22:
This pull request adds support for printing pseudoinstructions in rotor disassembly.

Pseudoinstructions are implemented (mostly) according to the [RISC-V spec](https://riscv.org//wp-content/uploads/2017/05/riscv-spec-v2.2.pdf), with a few exceptions noted by comments.

The aim of this pull request was to get Rotor to disassemble instructions as closely as possible to RISC-V objdump


ckirsch (changes requested):
@Jean-M22 looks really nice!

Please change the following (all easy to do):

- avoid changes to .gitignore
- change the type of rd, rs1, and rs2 to uint64_t, also in print_assembly, and then use get_register_name only in the last moment when printing
- add a flag for turning off printing of pseudo instructions, default should be on

Thanks a lot!

PR #383, opened 2024-05-22
Add missing compilable check to array target

SaTae66:
Although the [`local-declaration.c`](https://github.com/cksystemsteaching/selfie/blob/main/grader/assignments/array/local-declaration.c) exists in the grader, it is not used in the `[array]` grader target. I assume this unintentional, so here is the small fix. If this is indeed intentional, feel free to just close this PR.


ckirsch:
@SaTae66 thanks a lot!

PR #386, opened 2024-05-24
Upgrade Dockerfile to Ubuntu 23.04, GCC13, RISC-V zifencei extension

EETagent:
Let's see if the github action goes through

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -228,7 +228,7 @@ FROM selfieall AS selfieeverything
 
 # install tools for 32-bit selfie
 RUN apt-get update \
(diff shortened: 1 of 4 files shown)


ckirsch:
@EETagent log messages are out-of-order because of parallel make...the remaining issue is that the new gcc setup does not seem to compile 32-bit binaries. Look for rotor-32, for example, to see that. Any idea how to fix that?

EETagent:
@ckirsch ✅

ckirsch:
@EETagent thanks a lot!

PR #387, opened 2024-05-25
Only halt model evaluation on exit syscall

thomaswulz:
Rotor currently has a bug where single-instruction infinite loops are wrongly detected as exit syscalls:
```
0x100E8: addi a0,zero,42
0x100EC: jal zero,0x100EC <0>
./rotor: infinite-loop-rotorized.btor2 called exit(42) on core-0 @ 0x100EC after 2 steps
```

This happens because `eval_sequential` will indicate that the program has halted iff there is no state change after the execution of the current instruction. As far as I can tell, this only affects exit syscalls and trivial infinite loops like the one above. Therefore, I addressed this issue by changing the halting condition to a check if the current instruction is an exit syscall. With this change, such infinite loops do not get any special treatment and simply run until the timeout limit:

```
0x100EC: jal zero,0x100EC <0>
./rotor: terminating infinite-loop-rotorized.btor2 after 100000 steps
```

Alternatively, both of these checks could be combined to indicate whether the model evaluation was terminated due to a lack of state change or an actual exit syscall. However, I'm not sure if there is any practical use for detecting these trivial infinite loops in the model evaluation.

### DIFF
--- a/tools/rotor.c
+++ b/tools/rotor.c
@@ -12422,7 +12422,7 @@ void eval_multicore_states() {
 
     if (eval_multicore_sequential()) {
       if (number_of_cores > 1) {
-        printf("%s: %s called exit on all cores after %lu steps (up to %lu instructions)", selfie_name,
+        printf("%s: %s halted on all cores after %lu steps (up to %lu instructions)", selfie_name,
           model_name, current_step - current_offset, next_step - current_offset);
         if (any_input) printf(" with input %lu\n", current_input); else printf("\n");
       }
@@ -12467,7 +12467,7 @@ uint64_t eval_constant_propagation() {
   }
 
   if (eval_multicore_sequential()) {
-    printf("%s: %s called exit on all cores already after 1 step\n", selfie_name, model_name);
+    printf("%s: %s halted on all cores already after 1 step\n", selfie_name, model_name);
 
     return 1;
   }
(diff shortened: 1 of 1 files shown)


ckirsch:
@thomaswulz I believe there is a misunderstanding. Please keep the code as is, just generalize the message that indicates termination. In fact, you can distinguish there if termination happens because of no state change due to exit or due to a trivial infinite loop. You already have the code for that here which is nice!

thomaswulz:
I think the patch has the desired behavior now.

Infinite loop with unrolling:
```
./rotor: infinite-loop-rotorized.btor2 halted due to no state change on core-0 @ 0x100EC after 1 steps (up to 2 instructions)
./rotor: 70923 characters of model formulae written into infinite-loop-rotorized.btor2
./rotor: infinite-loop-rotorized.btor2 halted on all cores after 1 steps (up to 2 instructions)
```

Infinite loop without unrolling:
```
0x100E8: addi a0,zero,42
0x100EC: jal zero,0x100EC <0>
./rotor: infinite-loop-rotorized.btor2 halted due to no state change on core-0 @ 0x100EC after 1 steps (up to 2 instructions)
./rotor: executed up to 2 instructions without input
```

Exit with unrolling:
```
./rotor: asm-exit-with-one-rotorized.btor2 called exit(1) on core-0 @ 0x100F0 after 2 steps (up to 3 instructions)
./rotor: 71638 characters of model formulae written into asm-exit-with-one-rotorized.btor2
./rotor: asm-exit-with-one-rotorized.btor2 halted on all cores after 2 steps (up to 3 instructions)
```

Exit without unrolling:
```
0x100E8: addi a0,zero,1
0x100EC: addi a7,zero,93
0x100F0: ecall
./rotor: asm-exit-with-one-rotorized.btor2 called exit(1) on core-0 @ 0x100F0 after 2 steps (up to 3 instructions)
./rotor: executed up to 3 instructions without input
```

ckirsch:
@thomaswulz thanks a lot!

PR #388, opened 2024-05-30
feat: add option to grade a single commit link

nfejzic:
This PR implements an option in grader for grading of a single commit URL. 

I found this useful for cases where I wanted to double check a specific submission. Such example is: default compiler on macOS is `clang` even when invoking the `gcc` command. So there might be cases where I want to re-check with specific compiler (by setting the `CC` env variable accordingly), and grading a specific submission is very handy. 

Feel free to close the PR if you don't find this useful.

Example usage: 
`./grader/self.py -l https://github.com/some_repo/some_commit bitwise-and-or-not`

### DIFF
--- a/grader/lib/cli.py
+++ b/grader/lib/cli.py
@@ -31,6 +31,7 @@ For more detailed information about its usage, please consult the README.
 '''
 
 bulk_grade_mode = False
+commit_link_grade_mode = False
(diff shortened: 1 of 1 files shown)


ckirsch:
@nfejzic great! thank you!

PR #399, opened 2025-03-04
Update README.md

Kneidl18:
When preparing selfie for grading, I encountered the issue of not having an initialized repo. So here's the change to fix that


ckirsch:
@Kneidl18 thanks!

PR #405, opened 2025-04-02
Adding new bitwise-logical test

scrawfordlk:
To in particular test the correct operator precedence of & and |.


scrawfordlk:
I added a test for the assignment bitwise-and-or-not that checks whether the operator & has a higher precedence than the operator | (as it should).

ckirsch:
@scrawfordlk thanks a lot! This is simple enough to be included right now.

PR #411, opened 2025-06-10
Rename duplicated test method for config parsing

emmanuel-ferdman:
## PR Summary
This small PR fixes the renames the duplicated test method for config parsing.

### DIFF
--- a/tools/quarc/requirements.txt
+++ b/tools/quarc/requirements.txt
@@ -18,7 +18,7 @@ qiskit-aer==0.10.4
 qiskit-ibmq-provider==0.19.1
 qiskit-ignis==0.7.1
 qiskit-terra==0.20.2
-requests==2.32.2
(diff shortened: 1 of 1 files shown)


emmanuel-ferdman:
@ckirsch could you please review this small PR when you get the chance? thank you 🙌

emmanuel-ferdman:
@ckirsch The purpose of this change is to address an issue where there are two distinct test methods defined with the exact same name: `test_invalid_command_placement`. In Python, when you define multiple methods with the same name within a single class, the last definition overwrites any previous ones. As a result, only the second test (the one intended to check `invalid_format_cfg`) was actually being discovered and executed by the test runner. The first test (intended to check `invalid_command_placement_cfg`) was silent and never ran, potentially hiding a bug or a gap in the test coverage for that specific scenario.

PR #433, opened 2026-05-06
Split array assignment

scrawfordlk:
Ths splits the array assignment into two assignments:

### array-access
- Requires implementing array subscripting for pointers (assignment and use)

### array-allocation
- Requires implementing arrays as a supported data structure
- This is the same as the old array assignment


ckirsch:
@scrawfordlk very nice! Thanks a lot. This is definitely an improvement.

PR #435, opened 2026-05-07
Update book

scrawfordlk:
- Adds some more information on the new array assignments in `assignments/compiler-assignments.md`.
- Updates the book to discuss the following: 
    - heap-allocated arrays (as pointers to heap-allocated blocks of memory)
    - the first two array assignments and directions on how to solve them
- Does not update the PDF (as I am not aware of the tool used for the conversion)


ckirsch:
@scrawfordlk very nice again! Thank you so much!

PR #438, opened 2026-06-01
fix: local/global variable redefinitions now cause syntax errors

GGBOIH250:
fixed a TODO on line 4574 previously in `selfie.c` which mentions about checking local variable redefinitions as well. Now both local and global variables cannot be redefined similar to C.


ckirsch:
@GGBOIH250 just pushed a minor style update, after having Claude check impact on the rest of the system.

